Issue #55, opened 2025-08-20
Refactor spec for Gas Analyzer for block simulation approach and in general


tbsoc:
tx approach ain't working

we're throwing the table up to do whole block simulations

compare with @rubydusa secret repo

PR #56, opened 2025-08-24
Block approach

RonTuretzky:
Related to #55


rubydusa:
irrelevant because evmsketch

PR #78, opened 2025-12-16
Deterministic simulations

bagelface:
This PR adds changes that allow for more deterministic simulations.

- A user can specify and block height AND a block timestamp to anchor the anvil instance to.
- Logic updated to ensure only a single anvil is being used per GasKiller instance.

Additionally, we also:

- Refactor "GasKiller" to use a builder pattern for easier use and readability.
- Reenable previously "ignored" tests
- Run `cargo sort` to cleanup "Cargo.toml"
- Remove deprecated functions (no backwards compatibility)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,24 +13,24 @@ alloy = { version = "1.0.23", features = [
     "json-rpc",
 ] }
 alloy-dyn-abi = "1.0.23"
-alloy-rpc-types = { version = "1.0.23", features = ["trace"] }
+alloy-eips = "1.0.23"
 alloy-provider = { version = "1.0.23", features = ["debug-api"] }
-anyhow = "1.0.98"
-tokio = { version = "1.45.1", features = ["rt-multi-thread"] }
-url = "2.5.4"
-dotenv = "0.15.0"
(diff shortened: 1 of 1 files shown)


src/gk.rs
```diff
@@ -45,21 +54,86 @@ type ConnectHTTPDefaultProvider = FillProvider<
...
+            block.header.timestamp
+        } else {
+            let block = source_provider
+                .get_block_by_number(alloy::eips::BlockNumberOrTag::Latest)
+                .await?
+                .ok_or_else(|| anyhow!("Latest block not found"))?;
+            block.header.timestamp
+        };
+
+        // Determine the timestamp to use for the next block
```
RonTuretzky: what do you think the timestamp manipulations here are contributing?

bagelface: We emit a log when calling `sum` that includes the block timestamp https://github.com/BreadchainCoop/gas-killer-array-summation/blob/9ad682424ca12e4bc8f2819ab28a1d629213f147/src/ArraySummation.sol#L73

RonTuretzky (approved):

PR #96, opened 2026-02-28
feat: replay preceding block txs for accurate gas estimation

RonTuretzky:
## Summary

Closes #95

- Adds `PrecedingTx` struct and `replay_preceding_transactions()` to the `gas-estimator` crate — replays transactions via revm's `transact_commit` to advance `CacheDB` state
- Adds `get_preceding_transactions()` to the `rpc` crate — fetches block with full tx objects via a single `eth_getBlockByNumber(N, true)` call, converts to `PrecedingTx`
- Adds `estimate_state_changes_gas_with_preceding()` to `GasKillerEvmSketchDefault` — creates one `CacheDB`, replays preceding txs, then runs gas estimation on the same DB
- Wires it into the CLI: extracts `transaction_index` from receipt, fetches preceding txs when index > 0, gracefully falls back to current behavior on RPC failure

**Decisions (from spec):**
- Single RPC call (`eth_getBlockByNumber` with full tx objects) for fetching preceding txs
- Direct revm replay (not `debug_traceCall` per preceding tx)
- EIP-4788 beacon root handling is out of scope
- WASM path unchanged (no RPC access)

## Test plan

- [ ] Test with a first-in-block transaction (should skip replay, no behavior change)
- [ ] Test with a mid-block transaction (should replay preceding txs and show improved accuracy)
- [ ] Test with `--anvil` flag (Anvil path is unchanged, no regression)
- [ ] Verify graceful fallback when RPC doesn't support `eth_getBlockByNumber`


crates/cli/src/main.rs
```diff
@@ -296,9 +299,36 @@ async fn execute_command(cli_args: CliArgs) -> Result<()> {
...
+                    )
+                    .await
+                    .unwrap_or_else(|e| {
+                        eprintln!(
+                            "Warning: Failed to fetch preceding transactions: {}. \
+                             Estimating with block N-1 state.",
+                            e
+                        );
+                        Vec::new()
+                    });
```
rubydusa: should we really warn here? I think it's more appropriate to fail here

crates/evmsketch/src/lib.rs
```diff
@@ -231,6 +232,40 @@ impl GasKillerEvmSketchDefault {
...
+    ) -> Result<u64> {
+        let mut cache_db = CacheDB::new(&self.executor.sketch.rpc_db);
+        let gas_limit = self.executor.sketch.anchor.header().gas_limit;
+
+        if !preceding_txs.is_empty() {
+            gas_analyzer_estimator::replay_preceding_transactions(
+                &mut cache_db,
+                preceding_txs,
+                gas_limit,
+            )?;
```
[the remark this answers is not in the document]
rubydusa: not true - the header is the block of the simulated tx

crates/rpc/src/lib.rs
```diff
@@ -111,3 +113,60 @@ pub async fn compute_state_updates_from_tx<P: Provider + DebugApi>(
...
+
+    let txs: Vec<_> = block.transactions.into_transactions().collect();
+
+    if (tx_index as usize) > txs.len() {
+        bail!(
+            "Transaction index {} exceeds block transaction count {}",
+            tx_index,
+            txs.len()
+        );
+    }
```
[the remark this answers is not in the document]
rubydusa: addressed in [a8ce9a4](https://github.com/gas-killer/gas-analyzer/pull/96/commits/a8ce9a479c43890f382d09721b028e89ce208b96)

rubydusa:
## CI failure diagnosis & fix

The `test-native` failure on this PR is an **external regression**, not caused by anything on this branch. Pushed `c8e4495` to fix it.

### Symptom

```
test tests::test_compute_state_update_simulate_call ... FAILED

thread 'tests::test_compute_state_update_simulate_call' panicked at crates/core/src/trace.rs:38:45:
invalid hex: ParseIntError { kind: InvalidDigit }
```

### Why it's not the PR's fault

`git diff main..HEAD --stat` shows the PR touches `cli`, `evmsketch`, `gas-estimator`, `rpc`, and `wasm` — **not** `crates/core/src/trace.rs` or `crates/anvil/`. The failing test (`test_compute_state_update_simulate_call`) and the panicking parser (`parse_trace_memory`) are both untouched by this branch.

Branch CI history confirms this:

| Date | Foundry stable at the time | Result |
|---|---|---|
| 2026-04-02 (last green on this branch, after merging main) | v1.5.1 | ✅ |
| 2026-04-06 (last green on `main`) | v1.5.1 | ✅ |
| 2026-04-29 (this PR's CI runs) | **v1.7.0** (released 2026-04-28) | ❌ |

The CI workflow installs Foundry via `foundry-rs/foundry-toolchain@v1`, which always pulls the latest stable. The toolchain rolled from 1.5.1 → 1.7.0 between green and red, with no code change on our side.

### Root cause

Foundry v1.7.0 bumped `revm-inspectors` from `0.33.0` → `0.39.0`. Since [revm-inspectors v0.38.1](https://github.com/paradigmxyz/revm-inspectors/releases/tag/v0.38.1), [PR #425 (\"align memory encoding and returnData gating\")](https://github.com/paradigmxyz/revm-inspectors/pull/425) changed `convert_memory()`:

```diff
-        memory.push(hex::encode(chunk));
+        memory.push(hex::encode_prefixed(chunk));
```

So Anvil's `debug_traceTransaction` `StructLog.memory` entries now look like:

- Old: `"0000000000000000000000000000000000000000000000000000000000000001"` (64 hex chars, bare)
- New: `"0x0000000000000000000000000000000000000000000000000000000000000001"` (66 chars, prefixed)

The motivation for the upstream change is alignment with the [opcode tracer spec (execution-apis #762)](https://github.com/ethereum/execution-apis/pull/762).

Our `parse_trace_memory` (`crates/core/src/trace.rs:35`) joined the entries and parsed every 2 chars as a hex byte. With the new format, the second chunk is `\"0x\"`, and `u8::from_str_radix(\"0x\", 16)` returns `InvalidDigit` on the `'x'` — exactly the panic in the log. Geth/erigon and pre-1.7 Anvil still emit bare hex, so it's a forward-incompatibility, not a bug we introduced.

This affects only the Anvil backend; the EvmSketch path doesn't go through Anvil's structlogger, which is why `test-native`'s first step (`cargo test`, default = EvmSketch) passed and only the second step (`cargo test --features anvil`) failed.

### The fix (`c8e4495`)

In `parse_trace_memory`, strip an optional `0x` per entry before joining:

```rust
pub fn parse_trace_memory(memory: Vec<String>) -> Vec<u8> {
    memory
        .iter()
        .map(|s| s.strip_prefix(\"0x\").unwrap_or(s))
        .collect::<String>()
        .chars()
        .collect::<Vec<char>>()
        .chunks(2)
        .map(|c| c.iter().collect::<String>())
        .map(|s| u8::from_str_radix(&s, 16).expect(\"invalid hex\"))
        .collect::<Vec<u8>>()
}
```

This is backward-compatible: bare hex passes through `strip_prefix` unchanged, and prefixed hex gets the prefix removed. Both old and new Anvil, plus geth/erigon, now work.

I also added a `#[cfg(test)]` unit test in `crates/core/src/trace.rs` covering bare, prefixed, and mixed inputs. This locks in both formats so the same regression — or anything similar — gets caught in `core` without needing a live RPC + Anvil round-trip.

### Verification

- `cargo test -p gas-analyzer-core` ✅ (new unit test passes)
- `cargo clippy --workspace --exclude gas-killer-wasm --exclude gas-analyzer-anvil --all-targets -- -D warnings` ✅
- `cargo fmt --all -- --check` ✅
- The original `test_compute_state_update_simulate_call` integration test could not be reproduced locally (my Anvil is v1.5.1, pre-regression), but it goes through `parse_trace_memory` which now handles both formats. CI will confirm.

### Optional follow-up (not in this PR)

`foundry-rs/foundry-toolchain@v1` floats to latest stable, so any future Foundry release can break CI the same way. If we want determinism, pin the version:

```yaml
- uses: foundry-rs/foundry-toolchain@v1
  with:
    version: stable  # or a specific tag like v1.7.0
```

Happy to do that as a separate PR if desired.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/gas-estimator/src/lib.rs
```diff
@@ -195,3 +196,79 @@ where
...
+        let mut evm = Context::mainnet()
+            .with_db(&mut *cache_db)
+            .modify_cfg_chained(|cfg| {
+                cfg.disable_nonce_check = true;
+                cfg.disable_balance_check = true;
+                cfg.disable_base_fee = true;
+                cfg.disable_fee_charge = true;
+            })
+            .modify_block_chained(|block| {
+                block.basefee = 0;
```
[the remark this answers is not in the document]
rubydusa: Resolved in current code — `replay_preceding_transactions` now passes `number`, `timestamp`, `gas_limit`, `coinbase`, `prevrandao`, and `basefee` from `sim_env` (crates/gas-estimator/src/lib.rs:312-329).

crates/gas-estimator/src/lib.rs
```diff
@@ -195,3 +196,79 @@ where
...
+
+    for (i, tx) in preceding_txs.iter().enumerate() {
+        let mut evm = Context::mainnet()
+            .with_db(&mut *cache_db)
+            .modify_cfg_chained(|cfg| {
+                cfg.disable_nonce_check = true;
+                cfg.disable_balance_check = true;
+                cfg.disable_base_fee = true;
+                cfg.disable_fee_charge = true;
+            })
```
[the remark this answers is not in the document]
rubydusa: Resolved — doc rewritten to match the code: it now explicitly says fee transfers to coinbase are *not* applied because `disable_fee_charge=true` (crates/gas-estimator/src/lib.rs:295-299).

crates/gas-estimator/src/lib.rs
```diff
@@ -195,3 +196,79 @@ where
...
+            .modify_block_chained(|block| {
+                block.basefee = 0;
+                block.gas_limit = block_gas_limit;
+            })
+            .build_mainnet();
+
```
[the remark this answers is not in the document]
rubydusa: Resolved in 63ed6ec ("perf: build EVM once outside replay loop") — builder now lives above the `for tx in preceding_txs` loop.

crates/cli/src/main.rs
rubydusa: Done — current code uses `.map_err(...)?` to hard-fail on preceding-tx fetch errors instead of warning (crates/cli/src/main.rs:306-317).

crates/cli/src/main.rs
```diff
@@ -296,9 +299,37 @@ async fn execute_command(cli_args: CliArgs) -> Result<()> {
...
+                        block_number,
+                        tx_index,
+                    )
+                    .await
+                    .map_err(|e| {
+                        anyhow::Error::msg(format!(
+                            "Failed to fetch preceding transactions for block {} (tx index {}): {}",
+                            block_number, tx_index, e
+                        ))
+                    })?;
```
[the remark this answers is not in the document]
rubydusa: @RonTuretzky I'm against fallbacking here to previous behavior, even if the PR specifies it. WDYT?

crates/evmsketch/src/lib.rs
```diff
@@ -264,6 +266,42 @@ impl GasKillerEvmSketchDefault {
...
+    /// If `preceding_txs` is empty (first-in-block), this behaves identically
+    /// to `estimate_state_changes_gas`.
+    pub fn estimate_state_changes_gas_with_preceding(
+        &self,
+        contract_address: Address,
+        caller_address: Address,
+        state_updates: &[StateUpdate],
+        preceding_txs: &[PrecedingTx],
+    ) -> Result<u64> {
+        let mut cache_db = CacheDB::new(&self.executor.sketch.rpc_db);
```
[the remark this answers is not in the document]
rubydusa: addressed in this PR:
https://github.com/gas-killer/gas-analyzer/pull/122

I waited to merge it into this one to try to make this PR simpler to review - But I guess it does not make much sense to merge this to main with bugs

crates/cli/src/main.rs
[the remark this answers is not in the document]
bagelface: I agree falling back to previous behavior is not ideal because of silent incorrect result possibility. I can see adding a flag to allow it to force a result even if possibility of being inaccurate, but that seems like not very high value/priority at the moment.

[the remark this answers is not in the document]
rubydusa: Alright so marking this as resolved

crates/gas-estimator/src/lib.rs
```diff
@@ -174,14 +180,18 @@ where
             cfg.disable_balance_check = true;
             cfg.disable_base_fee = true;
             cfg.disable_fee_charge = true;
+            // Default is PRAGUE; use the newest spec revm knows so that
+            // contracts using post-Pectra opcodes (e.g. EOF in Fusaka/Osaka)
+            // don't halt with `NotActivated`.
+            cfg.spec = revm::primitives::hardfork::SpecId::OSAKA;
```
bagelface: Hardcoding `SpecId::OSAKA` applies Osaka gas rules and opcode availability to every block regardless of when it was mined. Analysis of any pre-Osaka historical transaction will use the wrong gas schedule — different SLOAD/SSTORE costs, different opcode availability — and may silently produce a wrong gas estimate. `SpecId` should be derived from the block number (or threaded in via `SimEnvOpts`) rather than fixed at the newest known fork.

PR #122 flagged this explicitly: *"Hard-coding SpecId::OSAKA misbehaves on pre-Osaka historical txs."* The same issue applies here and in `replay_preceding_transactions` below.

crates/gas-estimator/src/lib.rs
```diff
@@ -174,14 +180,18 @@ where
...
         })
         .modify_block_chained(|block| {
             block.number = U256::from(sim_env.number);
             block.timestamp = U256::from(sim_env.timestamp);
             block.gas_limit = sim_env.gas_limit;
             block.beneficiary = sim_env.coinbase;
             block.prevrandao = Some(sim_env.prevrandao);
-            block.basefee = 0;
+            block.basefee = sim_env.basefee;
             block.difficulty = U256::ZERO;
```
bagelface: `block.difficulty` is not in `SimEnvOpts` and is always hardcoded to zero. This is currently harmless for two reasons: `SpecId::OSAKA` routes the `DIFFICULTY`/`PREVRANDAO` opcode through `block.prevrandao` (EIP-4399), and post-Merge block headers already carry `difficulty = 0` by protocol.

However, once `SpecId` is correctly derived from block number (see line 186), pre-Merge blocks would use a pre-Paris spec where revm reads `block.difficulty` for the opcode — and would see 0 instead of the real PoW difficulty. Any contract using `DIFFICULTY` for entropy or access control on a pre-Merge block would execute incorrectly during replay.

Fix: add `difficulty: U256` to `SimEnvOpts`, populate it from `header.difficulty` in `sim_env()`, and use it here. The same issue exists in `replay_preceding_transactions` at line 336.

crates/gas-estimator/src/lib.rs
```diff
@@ -253,6 +263,101 @@ where
...
+/// opcode returns the same value the original tx observed.
+pub struct PrecedingTx {
+    pub from: Address,
+    pub kind: TxKind,
+    pub input: Bytes,
+    pub value: U256,
+    pub gas_limit: u64,
+    pub nonce: u64,
+    pub gas_price: u128,
+}
```
bagelface, line 327: `PrecedingTx` is missing two fields that affect whether replay produces the correct mid-block state:

**`access_list: AccessList`** (EIP-2930, active since Berlin): access lists pre-warm addresses and storage slots, reducing SLOAD from 2100→100 gas and cold account accesses from 2600→100. Without them, the same execution costs more gas during replay than in the original block. A preceding tx that set a tight gas limit relying on access list warming can OOG during replay. An OOG halt in `transact_commit` does **not** commit storage changes — only the nonce bump is applied — leaving the `CacheDB` missing those writes for everything that follows.

**`authorization_list`** (EIP-7702, active since Prague): set-code transactions atomically assign bytecode to EOAs. Without this field the replay skips code assignment, so any tx later in the block (including the one being analyzed) that calls into those EOAs sees an empty account instead of the delegated code.

Both fields should be added to the struct and populated from the alloy `Transaction` in `get_preceding_transactions`.

crates/gas-estimator/src/lib.rs
```diff
@@ -253,6 +263,101 @@ where
...
+    use revm::{ExecuteCommitEvm, MainBuilder, MainContext};
+
+    let mut evm = Context::mainnet()
+        .with_db(&mut *cache_db)
+        .modify_cfg_chained(|cfg| {
+            cfg.disable_nonce_check = true;
+            cfg.disable_balance_check = true;
+            cfg.disable_base_fee = true;
+            cfg.disable_fee_charge = true;
+            cfg.spec = revm::primitives::hardfork::SpecId::OSAKA;
```
bagelface: Same `SpecId::OSAKA` problem as line 186, but with a more direct correctness consequence here. In the replay path, a preceding transaction from a pre-Osaka block may behave differently under Osaka semantics: it might revert when it originally succeeded, or succeed when it originally reverted, resulting in different storage writes being committed to the `CacheDB`. Any subsequent read from those slots — including by the analyzed transaction's simulation — will see wrong state.

crates/gas-estimator/src/lib.rs
```diff
@@ -253,6 +263,101 @@ where
...
+        })
+        .modify_block_chained(|block| {
+            block.number = U256::from(sim_env.number);
+            block.timestamp = U256::from(sim_env.timestamp);
+            block.gas_limit = sim_env.gas_limit;
+            block.beneficiary = sim_env.coinbase;
+            block.prevrandao = Some(sim_env.prevrandao);
+            block.basefee = sim_env.basefee;
+            block.difficulty = U256::ZERO;
```
bagelface: Same latent `difficulty` bug as line 195 — hardcoded to zero, currently masked by `SpecId::OSAKA`. Needs to be fixed in tandem with correct `SpecId` derivation (line 327).

crates/gas-estimator/src/lib.rs
```diff
@@ -253,6 +263,101 @@ where
...
+        })
+        .build_mainnet();
+
+    for (i, tx) in preceding_txs.iter().enumerate() {
+        let revm_tx = TxEnv::builder()
+            .caller(tx.from)
+            .kind(tx.kind)
+            .data(tx.input.clone())
+            .value(tx.value)
+            .gas_limit(tx.gas_limit.min(EIP7825_TX_GAS_CAP))
```
bagelface: `EIP7825_TX_GAS_CAP` (2^24 = 16,777,216) is a per-tx gas limit introduced **only in Osaka**. On any earlier block a transaction can legitimately consume up to the block gas limit (~30M). If a preceding transaction originally succeeded with more than 16.7M gas used, capping its replay gas limit here causes an OOG halt mid-execution.

An OOG halt in `transact_commit` does **not** commit storage changes; only the sender's nonce is bumped. The `CacheDB` is then missing those storage writes, and the analyzed transaction's simulation observes stale pre-block data for every slot that preceding transaction would have written. The gas estimate silently degrades — or falls back to heuristic — without any indication that the replay state is wrong.

This cap should be removed from the replay path entirely (since `disable_fee_charge = true` makes the gas limit irrelevant to state correctness), or gated on `cfg.spec >= SpecId::OSAKA`.

(Note: the rubydusa comment on PR #122 dismissing this as "gas limit is taken from chain data, irrelevant" misses the point — the chain gas limit is taken, then *capped*, which can be less than what the transaction actually consumed.)

crates/evmsketch/src/lib.rs
```diff
@@ -264,6 +266,62 @@ impl GasKillerEvmSketchDefault {
...
+    /// Estimate gas for state changes, replaying preceding transactions first.
+    ///
+    /// Creates a single CacheDB, replays all preceding transactions to bring
+    /// it to the correct mid-block state, then runs gas estimation on that
+    /// state. This ensures the simulation sees the same state the original
+    /// transaction executed against.
+    ///
+    /// If `preceding_txs` is empty (first-in-block), this behaves identically
+    /// to `estimate_state_changes_gas`.
+    pub fn estimate_state_changes_gas_with_preceding(
```
bagelface, line 346: The public `estimate_state_changes_gas_raw` at line 142 was not updated by this PR. It still creates `CacheDB::new(&self.sketch.rpc_db)` — `BasicRpcDb` anchored to block N — so `eth_getStorageAt` calls return post-block state. This is the exact bug PR #122 fixed for `estimate_state_changes_gas` (line 246) and this function. Any external caller of `estimate_state_changes_gas_raw` gets subtly wrong results. It should be updated to use `SimpleRpcDb` at `block_number - 1`, consistent with its sibling methods.

bagelface (commented):
Correctness review. Six issues below, ranging from latent bugs to cases that actively corrupt the mid-block `CacheDB` state used for gas estimation.

crates/rpc/src/lib.rs
```diff
@@ -111,3 +113,62 @@ pub async fn compute_state_updates_from_tx<P: Provider + DebugApi>(
...
+                None => revm::primitives::TxKind::Create,
+            };
+            PrecedingTx {
+                from: tx.inner.signer(),
+                kind,
+                input: tx.inner.input().clone(),
+                value: tx.inner.value(),
+                gas_limit: tx.inner.gas_limit(),
+                nonce: tx.inner.nonce(),
+                gas_price: tx.inner.effective_gas_price(base_fee),
```
bagelface, line 178: `access_list` and `authorization_list` are not populated here (see comment on the `PrecedingTx` struct at line 287 of `gas-estimator/src/lib.rs`). Both are accessible on the alloy transaction: `tx.inner.access_list()` for EIP-2930 lists, and the authorization list for EIP-7702 type-4 transactions. Without them, replay accuracy degrades for any block containing access-list or set-code transactions.

bagelface (changes requested):
see comments

crates/gas-estimator/src/lib.rs
rubydusa: Addressed in 3e0f025. `SimEnvOpts` now carries a `spec: SpecId` field, and `EvmSketchExecutor::sim_env()` derives it from the anchored header via `alloy_evm::spec(&EthSpec::mainnet(), header)`. Both `estimate_gas_raw` and `replay_preceding_transactions` use `sim_env.spec` instead of a hardcoded `OSAKA`.

Tested by `test_sim_env_spec_derivation_against_mainnet` in `crates/evmsketch/src/lib.rs`, which feeds synthesized headers at known mainnet Berlin/London/Paris/Shanghai/Cancun heights through the same call site so an accidental chainspec swap (e.g. to Sepolia) surfaces as a wrong `SpecId`.

crates/gas-estimator/src/lib.rs
rubydusa: Addressed in 3e0f025. Added `difficulty: U256` to `SimEnvOpts`; `EvmSketchExecutor::sim_env()` populates it from `header.difficulty`, and both `estimate_gas_raw` and `replay_preceding_transactions` use `sim_env.difficulty` instead of `U256::ZERO`.

Tested by `test_difficulty_propagates_to_block_env` in `crates/gas-estimator/src/lib.rs`. It pre-deploys handwritten runtime `DIFFICULTY PUSH1 0 SSTORE` under `GRAY_GLACIER` (pre-Paris, where the opcode reads from `block.difficulty`), invokes it, and asserts slot 0 of the target equals `sim_env.difficulty`.

crates/gas-estimator/src/lib.rs
rubydusa: Addressed in 3e0f025. `PrecedingTx` gained `access_list: AccessList` and `authorization_list: Vec<SignedAuthorization>`. `replay_preceding_transactions` threads them into the revm `TxEnv` via `.access_list(...)` and `.authorization_list_signed(...)`. Population of these from the alloy `Transaction` happens in `gas_analyzer_rpc::get_preceding_transactions` (see the reply on the rpc/lib.rs comment).

Two tests in `crates/gas-estimator/src/lib.rs`:
- `test_access_list_threaded_through_replay` — replays the same SLOAD tx with vs without the slot pre-warmed and asserts a `gas_used` delta of exactly 2300 (EIP-2930 intrinsic 4300 minus warm-SLOAD savings 2000). A delta of zero would mean the field is being dropped.
- `test_authorization_list_applies_delegation` — signs an EIP-7702 authorization with a test private key, replays the tx, and asserts the authority's `CacheDB` account gains the 23-byte `0xef0100 || delegated_address` indicator.

crates/gas-estimator/src/lib.rs
rubydusa: Same fix as the comment on line 186 — `replay_preceding_transactions` now uses `sim_env.spec` (derived from the anchored header against `EthSpec::mainnet()`) instead of a hardcoded `OSAKA`. See 3e0f025 and `test_sim_env_spec_derivation_against_mainnet` in `crates/evmsketch/src/lib.rs`.

crates/gas-estimator/src/lib.rs
rubydusa: Same fix as the comment on line 195 — `replay_preceding_transactions` now uses `sim_env.difficulty` (populated from `header.difficulty` in `sim_env()`) instead of `U256::ZERO`. See 3e0f025 and `test_difficulty_propagates_to_block_env` in `crates/gas-estimator/src/lib.rs`.

crates/gas-estimator/src/lib.rs
rubydusa: Addressed in 3e0f025. The cap is removed from `replay_preceding_transactions` entirely. `estimate_gas_raw` still applies it but only via a new `effective_tx_gas_limit(gas_limit, spec)` helper that no-ops for pre-Osaka specs.

Two tests in `crates/gas-estimator/src/lib.rs`:
- `test_effective_tx_gas_limit_only_caps_under_osaka` — unit test on the helper; pre-Osaka inputs pass through unchanged, Osaka onward is capped at 2^24.
- `test_replay_does_not_apply_eip7825_cap_pre_osaka` — replays a `JUMPDEST/JUMP` gas-burner with `gas_limit = 25_000_000` under `SHANGHAI` and asserts the OOG `gas_used` exceeds 2^24, proving the limit was not truncated before reaching revm.

crates/evmsketch/src/lib.rs
rubydusa: Addressed in 3e0f025. `estimate_state_changes_gas_raw` now constructs `SimpleRpcDb { provider, block_number: anchor - 1 }` and wraps it in a `CacheDB`, mirroring the fix already applied to its sibling `estimate_state_changes_gas`.

Tested by `test_simple_rpc_db_queries_at_configured_block` in `crates/evmsketch/src/lib.rs`. It uses a custom recording `tower::Service` that captures every JSON-RPC request, then constructs a `SimpleRpcDb` with `block_number = 99` and asserts the resulting `eth_getStorageAt` call carries block tag `"0x63"` — so the underlying primitive cannot silently ignore its configured block.

crates/rpc/src/lib.rs
rubydusa: Addressed in 3e0f025. `get_preceding_transactions` now populates both fields from the alloy `Transaction`:

```rust
let access_list = tx.inner.access_list().cloned().unwrap_or_default();
let authorization_list = tx
    .inner
    .authorization_list()
    .map(<[_]>::to_vec)
    .unwrap_or_default();
```

Tested via `test_access_list_threaded_through_replay` and `test_authorization_list_applies_delegation` in `crates/gas-estimator/src/lib.rs` — same tests covering the `PrecedingTx` struct change.

crates/evmsketch/src/lib.rs
```diff
@@ -156,8 +166,12 @@ impl DefaultEvmSketchExecutor {
     /// `gas_price` defaults to 0 since it is a transaction-level field;
     /// callers with access to the original transaction can override it.
     /// `basefee` comes from the header (0 for pre-EIP-1559 blocks).
+    /// `spec` is derived from the header against Ethereum mainnet hardforks
+    /// (this binary is mainnet-only); `difficulty` carries the legacy PoW
+    /// value (zero post-Merge).
     pub fn sim_env(&self) -> SimEnvOpts {
         let header = self.sketch.anchor.header();
+        let spec = alloy_evm::spec(&alloy_evm::eth::spec::EthSpec::mainnet(), header);
```
bagelface: we can't assume mainnet spec. we will be running Sepolia transactions as well, so should detect from chain ID

bagelface (changes requested):
We don't want to assume mainnet specs, so need to detect Sepolia or Mainnet.

crates/evmsketch/src/lib.rs
rubydusa: Addressed in 6fdfdfc.

Chain detection now happens in `EvmSketchExecutorBuilder::build`: it queries `eth_chainId`, maps it through `chain_id_to_genesis_and_spec`, threads `Genesis::Mainnet | Genesis::Sepolia` into `EvmSketch::builder().with_genesis(...)`, and stores the chain ID on the executor. `sim_env()` then picks `EthSpec::mainnet()` or `EthSpec::sepolia()` based on the stored chain ID instead of hardcoding mainnet.

Unsupported chain IDs (Holesky, Anvil, etc.) error explicitly rather than silently defaulting to mainnet — defaulting would corrupt hardfork derivation whenever the target chain disagrees with mainnet at the same height/timestamp.

Two tests in `crates/evmsketch/src/lib.rs`:
- `test_chain_id_to_genesis_and_spec_supported_and_rejected_chains` — pins mainnet→`Genesis::Mainnet` + Cancun ts `1_710_338_135`, sepolia→`Genesis::Sepolia` + Cancun ts `1_706_655_072`, and asserts that 17000/31337/0 all error.
- `test_sepolia_spec_diverges_from_mainnet` — synthesizes a header at timestamp `1_708_000_000` (between the two Cancun activations) and asserts mainnet maps to `SHANGHAI` while Sepolia maps to `CANCUN`. Hardcoding mainnet would silently misclassify the Sepolia case as Shanghai.

bagelface (approved):
lgtm

PR #103, opened 2026-03-27
Fix decode_state_updates_tuple offset bug and improve test assertions

### LINKED ISSUE #73: Address issues from "state-updates-abi-encoding" PR review
## Context

The `state-updates-abi-encoding` branch was merged to unblock development, as it was a straightforward copy of an already working branch. However, a code review identified several issues that should be addressed in follow-up work.

### 1. Bug in `decode_state_updates_tuple` (Critical)

**Location:** `src/lib.rs` in `decode_state_updates_tuple` function

**Problem:** The offset calculation is incorrect. Offsets are relative to `head` (start of the offset array), but the code adds them to `tail`:

```rust
let start = tail + rel;  // WRONG: rel is relative to head, not tail
```

**Impact:** The decoder reads from incorrect positions in the buffer. Currently masked because tests only verify `data.len()`, not actual content.

**Fix:**
```rust
let start = head + rel;
```

### 2. Incomplete Test Assertions

**Location:** `src/lib.rs` tests `test_stateupdatetype_tuple_encoding` and `test_encoding_format`

**Problem:** Tests only verify `data.len() == 3` but don't check actual decoded content, which masked the decoding bug.

**Fix:** Add assertions that verify the actual decoded bytes match expected values.

### 3. Misplaced `#[allow(dead_code)]` Attribute

**Location:** `decode_state_updates_tuple` function

**Problem:** The function is marked `#[allow(dead_code)]` but it IS used in tests. This is misleading.

**Fix:** Either:
- Move the function inside `mod tests`, or
- Use `#[cfg(test)]` instead

### 4. Nested Helper Functions (Minor/Style)

**Location:** Helper functions `write_u256_word`, `pad32_len`, `encode_bytes`, `encode_bytes_array` defined inside `encode_state_updates_to_abi`

**Problem:** While functional, nested function definitions are unusual in Rust and reduce readability.

**Suggestion:** Consider extracting to private module-level functions.

## Priority

- Issue #1 (decoding bug) is critical if `decode_state_updates_tuple` will be used in production
- Issue #2 should be addressed alongside #1 to prevent regression
- Issues #3 and #4 are low priority cleanup

rubydusa:
## Summary
- Fix critical offset calculation bug in `decode_state_updates_tuple`: offsets are relative to `head`, not `tail` (`head + rel` instead of `tail + rel`)
- Add full round-trip assertions to `test_stateupdatetype_tuple_encoding` and `test_encoding_format` — decoded bytes are now compared against expected ABI-encoded values from `encode_state_updates_to_sol`
- Move `decode_state_updates_tuple` from `core/encoding.rs` into the anvil test module where it's actually used, removing the misleading `#[allow(dead_code)]`

Closes #73

## Test plan
- [x] `cargo test -p gas-analyzer-anvil -- test_stateupdatetype_tuple_encoding test_encoding_format` — both pass
- [x] `cargo check` — clean build, no warnings

### DIFF
--- a/crates/anvil/src/lib.rs
+++ b/crates/anvil/src/lib.rs
@@ -673,7 +673,10 @@ pub struct StateUpdateReport {
 mod tests {
     use super::*;
     use gas_analyzer_core::constants::*;
-    use gas_analyzer_core::{IStateUpdateTypes, StateUpdateType, decode_state_updates_tuple};
+    use gas_analyzer_core::{
+        IStateUpdateTypes, StateUpdateType, decode_state_updates_tuple,
+        encode_state_updates_to_sol,
+    };
 
     // Local sol! with #[sol(rpc)] for test that needs SimpleStorageInstance
     alloy::sol! {
@@ -709,7 +712,14 @@ mod tests {
             types,
             vec![U256::from(0u8), U256::from(2u8), U256::from(3u8),]
         );
-        assert_eq!(data.len(), 3);
+        let (_, expected_data) = encode_state_updates_to_sol(&state_updates);
+        assert_eq!(data.len(), expected_data.len());
+        for (i, (decoded, expected)) in data.iter().zip(expected_data.iter()).enumerate() {
+            assert_eq!(
+                decoded, expected,
+                "data[{i}] mismatch: decoded bytes don't match encoded input"
+            );
+        }
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


bagelface (approved):
The bug fix is correct — ABI `bytes[]` element offsets are relative to the start of the offset array (`head`), not past it, so `head + rel` is right and `tail + rel` was wrong.

Moving `decode_state_updates_tuple` into `mod tests` is the right call since it was only ever used in tests. The round-trip assertions are meaningful — comparing decoded output against `encode_state_updates_to_sol` actually validates the fix rather than just checking length.

Issue #4 from #73 (nested helpers in `encode_state_updates_to_abi`) is unaddressed but that was low priority and out of scope here.

LGTM.

PR #123, opened 2026-05-01
feat: add --verbose CLI mode for ABI decoding + traced rerun

rubydusa:
Adds an opt-in `-v` / `--verbose` mode for debugging failed estimations. Stacked on top of #122 — merge that first.

## What it does

When `--verbose` is passed and `ETHERSCAN_API_KEY` is set, three extra things happen at well-defined hook points; otherwise output is unchanged.

1. **Decoded state updates** — before estimation runs, every `StateUpdate::Call` is printed with its target's ABI-decoded function name, signature, and per-parameter values. Verified proxies are followed into their implementation ABI so proxied selectors decode (USDC `permit`/`transferFrom` etc.).

2. **Decoded RevertingContext** — when measured estimation fails with a `RevertingContext`, the failing call's args are decoded against the target's ABI and the revert data is decoded as `Error(string)`, `Panic(uint256)`, or any custom error in the ABI. Bytes from the trace stop being opaque.

3. **Traced rerun** — the failed estimation is re-executed under a revm `Inspector` that emits a per-frame call trace to stderr and flags the first frame that exits with empty returndata after consuming ≥31/32 of its forwarded gas. That signature is what both AllowanceHolder's `CheckCall`-via-`INVALID` and revm's `InstructionResult::NotActivated` produce on the parent — so the marker points directly at the originating frame instead of the bubbled-up outer revert. (This mode is what we used to identify Ekubo Core's `CLZ` opcode in the parent PR's debugging.)

## Modularization

```
crates/cli/src/verbose/
  mod.rs        - Verbose handle (no-op when disabled), public surface
  etherscan.rs  - Etherscan v2 client + per-address cache (HTTP only)
  decode.rs     - pure ABI decoders over raw bytes (no I/O)
  print.rs      - verbose printers built on top of the two above

crates/gas-estimator/src/
  tracing.rs    - CallTracer Inspector + estimate_state_changes_gas_traced
```

A single `Verbose` handle exposes async no-op-when-disabled methods so `main.rs` reads as a thin sequence of `verbose.print_state_updates(...)`, `verbose.print_reverting_context(...)`, `verbose.rerun_with_tracing(...)` — no `if verbose { ... }` walls.

`evmsketch` gains one thin `estimate_state_changes_gas_traced` wrapper that anchors at block N-1 via `SimpleRpcDb` and forwards into the gas-estimator's traced variant.

## Layering invariants

- `etherscan.rs` only does HTTP + caching, no formatting.
- `decode.rs` is pure: takes `&JsonAbi` and `&[u8]`, no I/O, no `println`.
- `print.rs` does all the formatting; only place that imports `colored`.
- The CLI never touches revm types directly — the inspector is fully contained in the gas-estimator crate.

## Help output

```
Flags:

  --anvil          Use Anvil-based implementation (requires --features anvil)
  --debug          Print full error details including RPC errors
  -v / --verbose   ABI-decode state updates and trace failed estimations
                   (requires ETHERSCAN_API_KEY environment variable)
```

## Test plan

- [x] Default (no `--verbose`) output unchanged
- [x] `--verbose` on a successful tx prints decoded state updates and the same final gas estimate
- [x] `--verbose` falls back gracefully if `ETHERSCAN_API_KEY` is missing (warning + run continues)
- [x] `--verbose` on a previously-failing tx (pre-Osaka-fix) printed the full call trace and pinpointed the `NotActivated`/empty-revert frame — that's how we found the `CLZ` issue in #122

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3263,7 +3263,10 @@ name = "gas-analyzer-cli"
 version = "0.1.0"
 dependencies = [
  "alloy",
+ "alloy-dyn-abi",
  "alloy-eips",
+ "alloy-json-abi",
+ "alloy-primitives",
  "alloy-provider",
  "anyhow",
  "colored",
@@ -3273,6 +3276,8 @@ dependencies = [
  "gas-analyzer-evmsketch",
  "gas-analyzer-rpc",
  "hex",
+ "reqwest",
+ "serde_json",
  "tokio",
  "url",
 ]
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -11,12 +11,17 @@ path = "src/main.rs"
 gas-analyzer-core = { path = "../core" }
 gas-analyzer-rpc = { path = "../rpc" }
 alloy = { version = "1.0.37", features = ["rpc", "rpc-types"] }
+alloy-dyn-abi = "1.0.37"
 alloy-eips = "1.0.37"
+alloy-json-abi = "1.0.37"
+alloy-primitives = "1.0.37"
 alloy-provider = { version = "1.0.37", features = ["debug-api"] }
 anyhow = "1.0.98"
 colored = "3.0.0"
 dotenv = "0.15.0"
 hex = "0.4"
+reqwest = { version = "0.12", features = ["json"] }
+serde_json = "1.0"
 tokio = { version = "1.45.1", features = ["rt-multi-thread"] }
 url = "2.5.4"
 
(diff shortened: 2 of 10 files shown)


rubydusa:
Converting this to draft as long as the previous pull requests are not merged, but this mode helped claude find two bugs - no support for Osaka opcodes like CLZ and a block minus one

rubydusa:
closing this. will redo this kind of feature after codebase cleanup
PR useful as a ref

PR #124, opened 2026-05-04
fix: forward original tx msg.value into state-update simulation

rubydusa:
## Summary

- Pass-through contracts (deposit-then-forward, intent settlers, swap routers) take ETH via `msg.value` and immediately forward it through a value-bearing CALL state update. The proxy invocation hard-coded `.value(ZERO)`, so the contract had no ETH to forward and the inner CALL halted with `OutOfFunds`, falling back to heuristic gas.
- Adds `SimEnvOpts.value`, makes `runStateUpdatesCall` payable, rebuilds the embedded artifact, and plumbs the original transaction's `value` through `GasKillerEvmSketch::estimate_state_changes_gas_with_preceding`. The CLI fetches it via `eth_getTransactionByHash`. The caller balance is topped up so revm's value transfer is well-defined under `disable_balance_check`.
- Pairs with [solidity-sdk: make verifyAndUpdate payable](https://github.com/gas-killer/solidity-sdk) so on-chain entry points accept the same forwarded value.

## Repro

Transaction `0xa5c2df3887e8576105f9169fb9d867d233539967b4c002fab3f158d49bcb0a81` was the original repro — the EOA sends 0.13 ETH to a router that pass-throughs the same 0.13 ETH to `0x0000000000001fF3684f28c67538d4D072C22734`.

Before:
```
GasKiller gas estimate: 536057 (heuristic - measured estimation failed)
```

After:
```
GasKiller gas estimate: 536136 (measured via StateChangeHandler)
```

## Test plan

- [x] `cargo test -p gas-analyzer-estimator -p gas-analyzer-core -p gas-analyzer-rpc -p gas-analyzer-evmsketch`
- [x] Re-ran the repro tx end-to-end — measured estimation succeeds
- [x] Re-ran a zero-value canonical tx (`0x9add9d…`) to confirm no regression on existing flows
- [x] CI on this branch


crates/gas-estimator/src/lib.rs
```diff
@@ -204,6 +209,28 @@ where
...
+            .flatten()
+            .unwrap_or_default();
+        cache_db.insert_account_info(
+            caller_address,
+            AccountInfo {
+                balance: caller_account.balance.saturating_add(sim_env.value),
+                nonce: caller_account.nonce,
+                code_hash: caller_account.code_hash,
+                code: caller_account.code,
+            },
```
bagelface, line 228: Use the struct-update shorthand instead of listing every field:

```rust
cache_db.insert_account_info(
    caller_address,
    AccountInfo {
        balance: caller_account.balance.saturating_add(sim_env.value),
        ..caller_account
    },
);
```

If revm ever adds a field to `AccountInfo`, the explicit literal silently drops it (and may not even fail to compile if the new field has `Default`). The spread avoids the problem entirely.

crates/cli/src/main.rs
```diff
@@ -133,6 +135,16 @@ async fn execute_command(cli_args: CliArgs) -> Result<()> {
...
 
+            // Fetch the original tx so we can mirror its `msg.value` during
+            // simulation. Pass-through contracts (deposit-then-forward, intent
+            // settlers, swap routers) lose ETH otherwise and value-bearing
+            // CALL state updates halt with OutOfFunds.
+            let tx_value: U256 = provider
+                .get_transaction_by_hash(bytes.into())
+                .await?
+                .map(|tx| tx.value())
+                .unwrap_or(U256::ZERO);
```
bagelface: We already hold a receipt for this hash, so `None` here means the node returned a receipt but then couldn't serve the transaction — an RPC inconsistency. Silently falling back to `U256::ZERO` re-introduces the original bug for that case without any signal. At minimum log a warning, or surface it as an error:

```rust
let tx = provider
    .get_transaction_by_hash(bytes.into())
    .await?
    .ok_or_else(|| anyhow!("tx {} present in receipt but missing from eth_getTransactionByHash", bytes))?;
let tx_value = tx.value();
```

crates/gas-estimator/src/lib.rs
```diff
@@ -46,15 +46,19 @@ pub(crate) fn effective_tx_gas_limit(gas_limit: u64, spec: SpecId) -> u64 {
...
 /// schedules, opcode availability, and per-tx limits, which can flip
 /// success/revert outcomes during preceding-tx replay and corrupt the
 /// CacheDB state the analyzed tx will read.
 ///
-/// `difficulty` is the legacy DIFFICULTY opcode value. Post-Merge it is zero
-/// by protocol; pre-Merge it carries real PoW difficulty and is read directly
-/// by the opcode under pre-Paris specs.
+/// `value` is the `msg.value` of the proxy invocation. Mirrors the original
+/// transaction's `value` so contracts that pass-through ETH (deposit-then-forward,
+///  intent settlers, swap routers) can fund value-bearing CALL state updates.
```
bagelface: Double space before "intent settlers" — copy-paste artifact.
`///  intent settlers` → `/// intent settlers`

bagelface (approved):
lgtm

crates/gas-estimator/src/lib.rs
bagelface: fixed ffc21f7b4a41b1b91f2fd99461e4de576e37165a

crates/gas-estimator/src/lib.rs
bagelface: fixed ffc21f7b4a41b1b91f2fd99461e4de576e37165a

crates/cli/src/main.rs
bagelface: fixed ffc21f7b4a41b1b91f2fd99461e4de576e37165a

PR #125, opened 2026-05-04
chore(submodule): retarget gas-killer-avs-sol to renamed solidity-sdk

rubydusa:
## Summary

- The submodule at \`contracts/lib/gas-killer-avs-sol\` was pointing at the old repo URL \`BreadchainCoop/gas-killer-avs-sol\`. The repo has since been renamed and transferred to \`gas-killer/solidity-sdk\`. GitHub still redirects, but the canonical URL belongs in \`.gitmodules\`.
- Bumps the pinned SHA from \`33ef63e\` to upstream \`main\` (\`ca0be70\`). The diff between the two SHAs is natspec-only on \`StateChangeHandlerLib.sol\` plus added unrelated contracts (ArraySummation factory) that we don't import.
- ABI of \`StateChangeHandlerGasEstimator\` is unchanged. Deployed bytecode is byte-identical except for the trailing IPFS metadata hash, so the embedded \`abis/StateChangeHandlerGasEstimator.json\` does not need to be refreshed.

## Test plan

- [x] \`forge build --force\` (contracts compile cleanly)
- [x] \`cargo build\` (workspace builds)
- [x] \`cargo test -p gas-analyzer-estimator\` (4/4 pass)
- [x] CI on this branch

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "contracts/lib/forge-std"]
 	path = contracts/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
-[submodule "contracts/lib/gas-killer-avs-sol"]
-	path = contracts/lib/gas-killer-avs-sol
+[submodule "contracts/lib/solidity-sdk"]
+	path = contracts/lib/solidity-sdk
 	url = https://github.com/gas-killer/solidity-sdk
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/foundry-rs/forge-std
 [submodule "contracts/lib/gas-killer-avs-sol"]
```
bagelface: would be nice to rename this to match the new submodule source name. "gas-killer-solidity-sdk"

bagelface (approved):
other than rename comment which isn't a blocker, this lgtm

bagelface (approved):
lgtm

PR #141, opened 2026-05-11
Use get proof when available, fallback to parallelized calls to get account details

### LINKED ISSUE #116: Batch/parallelize RPC calls in SimpleRpcDb
## Parent Epic
gas-killer/roadmap#9

## Context
\`crates/evmsketch/src/simple_rpc_db.rs\` implements revm's \`DatabaseRef\` trait. Each account read fires **sequential** async RPC calls bridged to sync via \`tokio::task::block_in_place()\`:

\`\`\`
basic_ref(address):
  get_balance()  →  block_in_place → RPC round-trip
  get_nonce()    →  block_in_place → RPC round-trip   ← sequential
  get_code_at()  →  block_in_place → RPC round-trip   ← sequential
\`\`\`

Each \`block_in_place\` blocks a worker thread. With a cold cache on a remote node, a single account lookup burns 3 serial round-trips (~25 ms each on mainnet). Measured in the investigation (#119): 266 distinct accounts during a full replay → 798 sequential round-trips just for account metadata.

## Relationship to #136

#136 proposes migrating \`SimpleRpcDb\` to \`eth_getProof\`, which reduces the per-account cost from 3 RPCs to 1 at the network level — strictly better than parallelizing. **However, \`eth_getProof\` is not universally supported** (some Anvil configurations, public RPC providers, and older archive nodes do not support it). #136 falls back to the 3-call path when the proof endpoint is unavailable.

This means the parallelization work here is still needed in two scenarios:
1. **Environments where \`eth_getProof\` is unsupported** — the fallback path is the only option, and it should not be left sequential
2. **As the implementation of the fallback path inside #136** — the fallback should use \`tokio::join!\` rather than 3 sequential \`block_in_place\` calls

Both issues should be worked on: #136 for the primary path, this issue for the fallback.

## Proposed Fix
Two complementary strategies — implement in order:

**1. Concurrent fetch on first access (low risk)**
Inside \`basic_ref()\`, fire \`get_balance\`, \`get_nonce\`, and \`get_code_at\` concurrently via \`tokio::join!\` inside a single \`block_in_place\`. Already-cached results skip the RPC call.

**2. Pre-warm cache before EVM execution (higher impact)**
Before calling \`evm.transact()\`, walk the tx access list (EIP-2930) and the known touched accounts from the trace, then batch-fetch all accounts in parallel before the EVM starts. Converts N×3 sequential round-trips into a single parallel fan-out.

## Files to Change
- \`crates/evmsketch/src/simple_rpc_db.rs\` — \`basic_ref()\`, add \`prefetch_accounts()\` method
- \`crates/evmsketch/src/lib.rs\` — call \`prefetch_accounts()\` before \`evm.transact()\`

## Acceptance Criteria
- Cold-cache account lookup fires at most 1 \`block_in_place\` per account (not 3)
- Pre-warm reduces total RPC round-trips for a typical 5-account tx by ≥50%
- Existing tests pass; new test verifies RPC call count via mock provider

## Unknowns / Risks
- \`DatabaseRef\` is \`&self\` (shared ref) — interior mutability required for the pre-warm cache; use \`RwLock\` or \`DashMap\`
- EIP-2930 access lists are not always populated — prefetch coverage varies by tx
- Public RPC providers may throttle parallel requests — need a configurable concurrency limit

### LINKED ISSUE #136: Reduce "SimpleRpcDb" account fetch from 3 sequential RPCs
## Parent Epic
gas-killer/roadmap#9

## Context
Extracted from the investigation in #119. \`SimpleRpcDb::basic_ref\` currently issues **3 sequential JSON-RPC calls** per cold account (\`eth_getBalance\`, \`eth_getTransactionCount\`, \`eth_getCode\`), each bridged to sync via \`tokio::task::block_in_place\`. For a tx that touches N cold accounts, that is 3N serialized round-trips at ~25 ms each.

From the investigation (\`trace_rpc_service\` example):
- ERC-20 transfer (3 state updates): estimate phase = **70% of hot-path** due to RPC fan-out
- UR execute (2 state updates): 30 simulate-RPCs issued from 2 top-level state updates because nested CALLs hit cold accounts on demand
- 266 distinct accounts in the historical replay run → 798 sequential round-trips just for account metadata

\`eth_getProof\` ([EIP-1186](https://eips.ethereum.org/EIPS/eip-1186)) returns balance, nonce, codeHash, and storage proofs in a **single call**. Migrating \`SimpleRpcDb\` to use \`eth_getProof\` as its primary fetch path cuts account reads from 3+K round-trips to 1 (where K is the number of storage slots for that account).

## Relationship to #116

#116 (parallelize \`basic_ref\` via \`tokio::join!\`) remains valid and should be worked alongside this issue:
- **When \`eth_getProof\` is supported**: this issue is the primary fix; #116's parallelization applies to the fallback path
- **When \`eth_getProof\` is unsupported**: #116 is the only fix available, and the fallback path here must use \`tokio::join!\` rather than 3 sequential calls

The fallback implementation in this issue **must** use the \`tokio::join!\` approach from #116 — not the current sequential path — so that unsupported-RPC environments still get a meaningful improvement.

## Approach

Replace \`basic_ref\` and \`storage_ref\` in \`crates/evmsketch/src/simple_rpc_db.rs\` with a \`eth_getProof\` call that fetches account metadata and requested storage slots together. When \`eth_getProof\` is unavailable, fall back to concurrent (not sequential) \`get_balance\`/\`get_nonce\`/\`get_code_at\` via \`tokio::join!\`.

```rust
// Rough shape of new basic_ref
async fn fetch_account_with_proof(
    provider: &RootProvider,
    address: Address,
    storage_keys: &[B256],
    block: BlockId,
) -> eyre::Result<(AccountInfo, HashMap<B256, U256>)> {
    let proof = provider
        .get_proof(address, storage_keys.to_vec())
        .block_id(block)
        .await?;

    // bytecode still needs a separate eth_getCode when code_hash != KECCAK_EMPTY
    // issue concurrently with the proof call via tokio::join!
    let code = if proof.code_hash != KECCAK_EMPTY {
        provider.get_code_at(address).block_id(block).await?
    } else {
        Bytes::new()
    };

    let account = AccountInfo {
        balance: proof.balance,
        nonce: proof.nonce.to::<u64>(),
        code_hash: proof.code_hash,
        code: Some(Bytecode::new_raw(code)),
    };

    let storage = proof.storage_proof
        .into_iter()
        .map(|sp| (sp.key.as_b256(), sp.value))
        .collect();

    Ok((account, storage))
}
```

**Note on storage keys**: The full set of storage slots accessed by a given account isn't known until \`transact()\` runs. A first-pass prefetch can use the tx's EIP-2930 access list + slots observed in the trace. Cold-miss slots encountered during simulation still fall back to individual \`eth_getStorageAt\` calls — but the common-case account metadata fetch drops from 3 RPCs to 1.

## Files to Change
- \`crates/evmsketch/src/simple_rpc_db.rs\` — replace \`basic_ref\`, \`storage_ref\` implementations; add fallback guard using \`tokio::join!\` (#116 approach)
- \`crates/evmsketch/src/lib.rs\` — update prefetch/inject callers if the interface changes

## Acceptance Criteria
- Cold EOA lookup fires exactly 1 RPC (\`eth_getProof\`) when supported; 3 concurrent RPCs (\`tokio::join!\`) when not
- Cold contract account fires at most 2 RPCs (\`eth_getProof\` + \`eth_getCode\`, concurrently) when supported
- On \`eth_getProof\` failure, falls back to the concurrent 3-call path from #116 without surfacing an error to the caller
- All existing tests pass
- RPC call count in the criterion/example benchmark drops by ≥50% for a cold-cache tx with multiple touched accounts

## Unknowns / Risks
- Not all RPC providers support \`eth_getProof\` (some Anvil configurations, older archive nodes); fallback path must be reliable
- The storage key set needed per account is not fully known pre-execution — prefetch coverage depends on access list quality
- Alloy's \`get_proof\` builder should be verified against mainnet providers before merging

bagelface:
## Summary

Closes #136 (partially — storage slot prefetch is tracked separately in #142).

Replaces `SimpleRpcDb::basic_ref`'s 3 sequential RPC calls (`eth_getBalance`, `eth_getTransactionCount`, `eth_getCode`) with a two-tier strategy:

**Primary path — `eth_getProof` (EIP-1186):**
- `eth_getProof` and `eth_getCode` are issued concurrently via `tokio::join!`
- EOAs: 1 RPC total (was 3 sequential)
- Contracts: 2 concurrent RPCs (was 3 sequential)

**Fallback path — parallel calls (fixes #116):**
- Triggered only when the RPC responds with an explicit `-32601` "Method not found" error; transient failures (network timeouts, connection resets) propagate to the caller and do not affect the fallback state
- Once detected as unsupported, an `AtomicBool` skips `eth_getProof` for all subsequent calls in the same `SimpleRpcDb` instance
- `eth_getBalance`, `eth_getTransactionCount`, and `eth_getCode` are issued concurrently via `tokio::join!` — latency drops from 3×RTT to ~1×RTT

Both paths reduce per-account latency from 3 serialised round-trips to approximately 1 round-trip. The proof path additionally cuts the total RPC count by 3×, which matters for provider rate limits and non-HTTP/2 connections.

## Files Changed

- `crates/evmsketch/src/simple_rpc_db.rs` — rewrote `basic_ref`; added `try_fetch_with_proof`, `fetch_account_parallel`, `GetProofOutcome`, `is_method_not_found`, and `SimpleRpcDb::new` constructor
- `crates/evmsketch/src/lib.rs` — updated all `SimpleRpcDb` construction sites to use `::new()`
- `crates/evmsketch/benches/fixtures/generate_replay_fixture.rs` — same constructor update

## Benchmark

CI offline benches show no change (expected — they use pre-populated `CacheDB<EmptyDB>` and never call `basic_ref`). The improvement is only visible in the live RPC path.

Local `end_to_end` bench against the configured QuikNode endpoint (which **does** support `eth_getProof`):

| metric | `main` | this branch | change |
|--------|:-------:|:-----------:|:------:|
| wall time (median) | 2.71 s | 2.57 s | **−5.3%** |
| CI range | — | [2.51 s, 2.68 s] | |

The pinned tx touches few cold accounts so the absolute saving is small. Txs touching many cold accounts (e.g. the 266-account replay in #136) would compound the saving proportionally.

### DIFF
--- a/crates/evmsketch/benches/fixtures/generate_replay_fixture.rs
+++ b/crates/evmsketch/benches/fixtures/generate_replay_fixture.rs
@@ -152,10 +152,7 @@ async fn run(rpc_url: String) -> Result<()> {
     // from the chain on demand.  After replay the cache keys tell us exactly
     // which accounts and slots need to be in the offline fixture.
     let state_block = block_number.saturating_sub(1);
-    let simple_db = SimpleRpcDb {
-        provider: any_provider,
-        block_number: state_block,
-    };
+    let simple_db = SimpleRpcDb::new(any_provider, state_block);
     let mut cache_db = CacheDB::new(simple_db);
 
     eprintln!("Replaying {tx_index} preceding txs to warm the cache (makes RPC calls)...");
--- a/crates/evmsketch/src/lib.rs
+++ b/crates/evmsketch/src/lib.rs
@@ -188,10 +188,7 @@ impl DefaultEvmSketchExecutor {
         gas_price: u128,
     ) -> Result<u64> {
         let state_block = self.anchor_block_number().saturating_sub(1);
-        let simple_db = SimpleRpcDb {
-            provider: self.sketch.provider.clone(),
-            block_number: state_block,
-        };
+        let simple_db = SimpleRpcDb::new(self.sketch.provider.clone(), state_block);
         let mut cache_db = CacheDB::new(simple_db);
         let mut sim_env = self.sim_env();
         sim_env.gas_price = gas_price;
@@ -322,10 +319,7 @@ impl GasKillerEvmSketchDefault {
     ) -> Result<u64> {
         // Use block_number - 1 (pre-transaction state), matching the Anvil path.
         let state_block = self.executor.anchor_block_number().saturating_sub(1);
-        let simple_db = SimpleRpcDb {
-            provider: self.executor.sketch.provider.clone(),
-            block_number: state_block,
-        };
+        let simple_db = SimpleRpcDb::new(self.executor.sketch.provider.clone(), state_block);
         let mut cache_db = CacheDB::new(simple_db);
         let sim_env = self.executor.sim_env();
         gas_analyzer_estimator::estimate_state_changes_gas(
@@ -370,10 +364,7 @@ impl GasKillerEvmSketchDefault {
         // tx, causing signature recovery to mismatch and the call to revert
         // with "invalid signature".
         let state_block = self.executor.anchor_block_number().saturating_sub(1);
-        let simple_db = SimpleRpcDb {
-            provider: self.executor.sketch.provider.clone(),
-            block_number: state_block,
-        };
+        let simple_db = SimpleRpcDb::new(self.executor.sketch.provider.clone(), state_block);
         let mut cache_db = CacheDB::new(simple_db);
         let mut sim_env = self.executor.sim_env();
 
@@ -639,10 +630,7 @@ mod tests {
         let client = RpcClient::new(transport, true);
         let provider: RootProvider<AnyNetwork> = RootProvider::new(client);
 
-        let db = SimpleRpcDb {
-            provider,
-            block_number: 99,
-        };
+        let db = SimpleRpcDb::new(provider, 99);
 
         // SimpleRpcDb's storage_ref blocks the current thread; spawn_blocking
         // gives it the worker thread it needs under multi_thread runtime.
--- a/crates/evmsketch/src/simple_rpc_db.rs
+++ b/crates/evmsketch/src/simple_rpc_db.rs
@@ -1,8 +1,11 @@
-use alloy::primitives::{Address, B256, U256};
+use std::sync::atomic::{AtomicBool, Ordering};
+
+use alloy::primitives::{Address, B256, Bytes, U256};
 use alloy::providers::Provider;
 use alloy_provider::RootProvider;
 use alloy_provider::network::AnyNetwork;
 use revm::database_interface::{DBErrorMarker, DatabaseRef};
+use revm::primitives::KECCAK_EMPTY;
 use revm::state::{AccountInfo, Bytecode};
 
 /// Error type for [`SimpleRpcDb`].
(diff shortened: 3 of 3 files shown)


bagelface:
## Local `end_to_end` Benchmark

> CI excludes this bench (network latency); run manually with `make bench-rpc`.

**Machine:** Intel Core i5-8279U @ 2.40GHz, macOS 15.7.5, rustc 1.94.0
**Pinned tx:** `0x680e2abfbccaf6246b4bda0989fc55dee169d0f6aef2ca4c63a17c6a8a39d6cb`

| metric | `main` (baseline) | this branch | change |
|--------|:-------------------:|:-----------:|:------:|
| wall time (median) | 2.71 s | 2.57 s | **−5.3%** |
| CI range | — | [2.51 s, 2.68 s] | |

The upper bound of the new range (2.68 s) sits below the baseline median (2.71 s), so the improvement is real rather than noise.

The gain is modest because the pinned tx touches few cold accounts — the 3N → N RPC reduction compounds more dramatically on txs with many cold accounts (e.g. the 266-account replay case from #136).

rubydusa:
eth_getProof is an interesting hack for this, I want to point out that JSON-2.0 RPC spec supports batching. It would be interesting to compare eth_getProof and batching the information needed, but given the fact that most of the hot path is debug_traceCall I wouldn't bother with it now

rubydusa:
I don't like the approach of falling back to the old behavior because it creates more code and paths to maintain and I think it'd be smarter to just replace existing functionality with eth_getProof.
OR at the very least separate them to two different structs - EthProofRpcDb and SimpleRpcDb, one with the old behavior and the other with the new. WDYT?

bagelface:
> I don't like the approach of falling back to the old behavior because it creates more code and paths to maintain and I think it'd be smarter to just replace existing functionality with eth_getProof. OR at the very least separate them to two different structs - EthProofRpcDb and SimpleRpcDb, one with the old behavior and the other with the new. WDYT?

I think it's important to keep the fallback. We orginally were using `eth_getProof` but switched because some RPCs didn't support it or the proofs weren't available if you went far back enough (can't remember which it was). I'll try splitting them into different structs to see how that feels, but it may end up being more code to maintain that way.

PR #163, opened 2026-06-26
Hybrid prestate/struct-log extraction (survive heavy-compute calls)

hudsonhrh:
## Problem
`call_to_encoded_state_updates_with_evmsketch` extracts diffs only from the struct-log trace (`compute_state_updates`), which is `O(execution steps)`. Heavy-compute tracked functions time out the node (`debug_traceCall` returns `-32000 execution timeout`) even when the resulting diff is tiny — e.g. an on-chain Game-of-Life `step` is ~16.8M gas with a multi-hundred-MB trace, yet only a 3-op diff. Those calls can't be served today.

## Change
Add a cheap **prestate fast path** (`prestateTracer` diffMode + `callTracer` logs, `O(changed slots)`) and dispatch to it when an eligibility check proves it sound, otherwise fall back to the existing struct-log path. Same STORE/CALL/LOG op model; **never emits an unsound diff**.

- **`crates/core`** (pure, 15 unit tests): `classify_prestate_eligibility` mirrors the struct-log replay model — DELEGATECALL/CALLCODE are transparent, while a regular CALL, CREATE/CREATE2/SELFDESTRUCT, or any non-consumer storage change forces fallback (a net diff can't separate the consumer's top-level writes from sub-call-induced ones). `build_state_updates_from_prestate` emits slot-sorted STOREs and logs in emission order (interleaved by `position`, with the global `index` as an authoritative override).
- **`crates/rpc`**: `get_prestate_diff_from_call`, `get_call_frame_from_call`.
- **`crates/evmsketch`**: try the fast path; fall back on ineligible calls *or* nodes that don't support the tracers — non-regressive.

## Compatibility
No public API change — the function dispatches internally, so the service picks this up on a dependency bump with zero code changes. Eligible heavy-compute calls now extract successfully; light calls get a cheaper, net-deduplicated extraction.

## Testing
15 new `core` unit tests cover storage diffing (changed/zeroing/net-zero/empty/consumer-only), LOG0–4 mapping, delegatecall-transparent log collection, position/index ordering, and the classifier (eligible vs fallback for regular CALL, CREATE, CALL-nested-in-delegatecall, cross-contract storage, balance-only). Full `gas-analyzer-core` suite (18) green; `gas-analyzer-rpc` and `gas-analyzer-evmsketch` type-check; `cargo fmt --check` clean. The approach was validated end-to-end (extract → apply via `verifyAndUpdate`, both regimes) in a separate reference harness, including an adversarial soundness audit that caught and fixed a delegatecall log-ordering case (now `position`-interleaved).


BreadrichEngels (changes requested):
## Review: hybrid prestate/struct-log extraction

The engineering here is genuinely good — the classifier faithfully mirrors the struct-log replay model, the log-order reconstruction (position-interleave + global-index override, reverted-frame exclusion) is careful, the eligibility gates are sound for *end state*, and the 15 unit tests are thorough. The ~940× win on heavy-compute calls is real and valuable. But I'm requesting changes on one thing the soundness argument doesn't address: **the extracted `state_updates` are digest-affecting, and the fast path produces a different *representation* than the struct-log paths — so this can make two honest provers sign different digests for the same call.**

### Why this is a consensus problem, not just a perf change
`state_updates` → `encode_state_updates_to_abi` → `storage_updates`, and `storage_updates` is hashed into the signed task digest (`sha256(transitionIndex, address(this), …, storageUpdates)`), which every validator recomputes and must match. So the digest is over the *representation*, not just the resulting state. "Applying the fast-path diff reproduces the same final state and events" (true) is **not** the same as "produces the same digest" (false).

**Concrete, certain divergence under `Legacy` (the encoding the service signs by default, per service#332):** a slot written twice in one call — say `X` then `Y` — yields
- struct-log Legacy: `[Store(slot, X), Store(slot, Y)]` (per-write, trace order)
- prestate fast path: `[Store(slot, Y)]` (net-deduplicated)

Different vectors → different ABI bytes → different digest. Your own writeup confirms this ("net-deduplicated … lower gas estimate than the struct-log paths report"). There are further gaps even against `Canonical` (e.g. a slot written then reverted to its original value: the canonical encoder re-asserts `Store(slot, original)` from its journal, but `prestateTracer` diffMode omits a net-zero slot entirely, so the fast path drops it), plus the reconstructed log order is a third byte-identity surface. But the Legacy net-dedup case alone is enough.

### Why the dispatch design makes it reachable
1. **Path selection is per-node, not coordinated.** `extract_state_updates_hybrid` prefers the fast path and falls back on *tracer error* (a node whose RPC lacks `prestateTracer`/`callTracer`) or ineligibility. So node A (tracers supported) signs the net-deduped digest while node B (no tracer support, or the call classified differently) recomputes the struct-log digest — they disagree, and the honest signature is rejected. For heavy-compute calls the struct-log path times out, so *those* effectively force every node onto the fast path (fine, homogeneous) — but for **light eligible calls** the fast path is taken *preferentially* even though struct-log would succeed, and that's where a mixed fleet forks.
2. **It's taken under `Legacy` too.** The fast path runs before the `match encoding`, so setting `STATE_ENCODING=Legacy` no longer yields the legacy representation for eligible calls — it silently yields the net form.
3. **"Picks this up on a dependency bump with zero code changes"** is exactly the unsafe rollout: bumping gas-analyzer in the service silently changes the digest for every eligible call, and a non-atomic bump (node ahead of validators, or mixed versions mid-deploy) forks consensus. This is the "the encoder version is consensus-critical → fleet-wide atomic rollout" property I flagged on #173/#332 — but here it's not even behind a config flag.

### What I'd want before this can be consumed by the service
Treat the extraction representation as consensus-critical and make it deterministic across all provers for a given call. Concretely, any one of:
- **Only engage the fast path when the struct-log path is actually infeasible** (i.e. as a timeout fallback for heavy-compute calls), rather than preferentially for light calls — so a call that *can* be served by the signed encoding always is. This shrinks the fast path to the case where every node must use it anyway.
- **Or** make the fast path byte-identical to the active encoding's output. That's impossible for `Legacy` (net-dedup vs per-write) and lossy for `Canonical` (the net-zero-written-slot case `prestateTracer` can't see), so this likely means redefining the signed digest to be over a *canonicalized net diff* that both the fast path and the struct-log encoders normalize to identically.
- **Or** gate it behind the same explicit, coordinated flag as `StateEncoding`, documented as requiring guaranteed tracer support fleet-wide + atomic rollout, and never mixed with a struct-log fallback for the same call.

You clearly already thought about path-equivalence — the reverted-top-level-call fallback (empty net diff vs struct-log emitting the rolled-back writes) is exactly this reasoning applied to one case. The ask is to extend that same rigor from "same end state" to "same signed representation," because that's what the digest commits to.

### Non-blocking notes
- The gas *estimate* also shifts (net-dedup → fewer stores), but that's discarded downstream (service#332), so it's only a reporting change — not a concern beyond the digest.
- `skipped_opcodes` always empty on the fast path is fine as documented (transient storage can't be observed given eligibility rules out external calls).
- Nice touch pinning both paths to the same update count in the bench's sanity assert — consider strengthening that to assert *vector equality* under Canonical, which would surface the net-zero-written-slot gap above.

Happy to re-review once the fast path is scoped so all provers deterministically agree on the signed `storage_updates` for a given call. The core extraction logic is solid; it's the automatic, per-node, under-any-encoding dispatch into a digest-affecting representation that needs to change.

bagelface:
You were right, and the "same end state ≠ same digest" framing was the thing to push on. Fixed by making the net diff a **distinct signed encoding** rather than an optimisation applied underneath the others.

## `StateEncoding::PrestateNet`

The three concerns map to three properties:

1. **It can't run under another encoding.** `STATE_ENCODING=legacy` yields the legacy representation again — the prestate tracers aren't even issued under `Legacy`/`Canonical`, so their latency goes away too.
2. **A tracer failure is a hard error, not a downgrade.** A node whose RPC lacks `prestateTracer` now fails the task rather than silently signing a different representation. Failing is recoverable; a divergent digest isn't.
3. **Not a per-node decision.** Selecting the representation is a deployment-wide config choice on the same knob as `Canonical`, with the same atomic-rollout discipline you set out in service#334 constraint 1.

Calls with no net form fall back to the **canonical** encoder within `PrestateNet`, so both of its representations stay revert-aware and slot-sorted. The classifier deciding which is which reads only the call tree and the diff — both fixed by the call and the block — and checks `error.is_some()` rather than error strings, so it doesn't fork across node implementations.

Also worth stating plainly: **merging this is a no-op for production.** Default stays `Legacy`, and `PrestateNet` isn't merely disabled service-side — `parse_state_encoding` accepts only `legacy|canonical` and panics otherwise, so a node can't start with it even by mistake. It becomes reachable only when someone writes the service-side plumbing, which sequences behind service#334's Helm work.

## On the non-blocking notes

Gas estimate and `skipped_opcodes` — agreed, left as-is.

The bench assert I did **not** do as suggested, and I think the suggestion backfires here. Vector equality would *pass* on that workload: 16 slots, each written once, in ascending slot order, with a single trailing log. That's exactly the narrow shape where the encoders agree, so the assert would read as proof of a general equivalence that doesn't hold. Instead:

- The divergences are pinned as a `core` test, `net_form_omits_slots_the_canonical_encoder_re_asserts`, covering both the repeated-write collapse and the net-zero-slot gap you identified.
- The bench assert stays a fixture guard with its scope stated, and the module docs now say it measures cost rather than equivalence.

Happy to switch to a literal vector-equality assert on a workload built to diverge if you'd rather have it in the bench.

## One thing that came out of this, re: the cutover

Chasing the net-zero case answered service#334's open question, and the answer isn't the convenient one — **canonical isn't digest-equivalent to legacy for no-external-call targets either**. Canonical journals stores and flushes one slot-sorted slice at the end, while legacy interleaves in execution order, so:

| target does | legacy | canonical |
|---|---|---|
| write a slot, then emit an event | `[Store, Log]` | `[Log, Store]` |
| write the same slot twice | both writes | last write only |
| write slot `BB`, then slot `AA` | write order | slot order |

They agree only when each slot is written once, in ascending order, with no log before the last write. Details in [service#334](https://github.com/gas-killer/service/issues/334#issuecomment-5203299469) — flagging it here because it means the `Legacy → Canonical` cutover is already a coordinated digest migration, so adding a third encoding doesn't introduce a new class of rollout risk, it joins one that exists.

The broader "make the encodings reconcilable so this stops being consensus-critical config" work is gas-analyzer#178, including your log-`index`-vs-`position` point as a second byte-identity surface.

BreadrichEngels (approved):
## Follow-up review (head `ff94dc2`) — consensus concern resolved ✅

This is exactly the right fix, and cleaner than any of the three options I sketched: the net form is now a **distinct `StateEncoding::PrestateNet`**, not an optimisation applied underneath the others. All three failure modes I raised are closed, and I verified each in the `evmsketch` dispatch:

1. **It can't run under another encoding.** `extract_state_updates_hybrid` only issues the tracers when `encoding.signs_prestate_net()`, which is true solely for `PrestateNet` — so `Legacy`/`Canonical` go straight to the struct-log path and yield their original representation (and shed the tracer latency). The `only_prestate_net_encoding_signs_the_net_form` test pins it.
2. **A tracer failure is a hard error, not a silent downgrade.** `try_prestate_net(...).await?` propagates the error out of the extractor rather than falling through to struct-log, so a node whose RPC lacks `prestateTracer`/`callTracer` fails the task instead of signing a different representation. That's the correct trade: a failed task is recoverable, a divergent digest isn't.
3. **Not a per-node decision.** It's a deployment-wide encoding on the same knob as `Canonical`, and it's genuinely unreachable in production — the service's `parse_state_encoding` accepts only `legacy|canonical` and panics otherwise, `default_encoding_is_legacy` guards the default, so **merging is a no-op for every current deployment.** It only becomes selectable once service#334 plumbs it.

The supporting decisions are right too: pairing `PrestateNet`'s struct-log fallback with `Canonical` (not `Legacy`) keeps both of its representations revert-aware and slot-sorted; deciding eligibility from the call tree + diff and gating on `error.is_some()` (not error *strings*) keeps the verdict from forking across client implementations; and the type-level doc now states plainly that the encoding selects the *signed representation* and is consensus-critical — exactly the framing that was missing. Concurrent tracer issuance is a nice touch given each re-simulates.

**On the bench assert: you were right and I was wrong.** A vector-equality assert on that fixture would *pass* — 16 single-write slots in ascending order with one trailing log is precisely the shape where the encoders agree — so it would read as proof of a general equivalence that doesn't hold. The dedicated `net_form_omits_slots_the_canonical_encoder_re_asserts` core test pinning the repeated-write collapse and the net-zero-slot gap is the better guard, and keeping the bench scoped to *cost* with that stated is the honest call.

### The finding worth carrying forward
Your point that **`Canonical` isn't digest-equivalent to `Legacy` either** (write-then-log interleave vs slot-sorted flush, repeated-write collapse, slot vs execution order) is the important generalisation here: the `Legacy → Canonical` cutover is *already* a coordinated digest migration, so `PrestateNet` joins an existing rollout-risk class rather than creating a new one. Good that it's captured in service#334 (atomic rollout) and gas-analyzer#178 (reconciling the encodings so this stops being consensus-critical config).

### One residual for the service#334 enablement (not this PR)
`PrestateNet`'s determinism now rests, beyond the classifier, on **cross-node equivalence of the `callTracer`/`prestateTracer` output itself** — the call-tree shape and especially the log `position`/`index` fields the ordering reconstruction depends on. `prestateTracer` diffMode is the raw state diff (deterministic across clients), but `callTracer` log ordering is a less battle-tested byte-identity surface than struct-log traces, and it's exactly the second surface gas-analyzer#178 already lists. So before `PrestateNet` is turned on fleet-wide, I'd want that cross-client (or pinned-single-tracer) equivalence nailed down the same way. Purely forward-looking; nothing to do here.

Clearing my CHANGES_REQUESTED. Nicely handled — the "distinct signed encoding" framing is the correct mental model, and the tests encode it.

PR #164, opened 2026-07-01
feat: report gas estimates for both BLS and Schnorr signature schemes

bagelface:
## Summary

The gas analyzer previously applied a single hard-coded Turetzky upper gas limit (the GasKiller gas floor). That floor depends on the signature scheme used to verify the aggregated operator attestation on-chain, so this reports figures for **both** schemes:

- **BLS** — `250000`
- **Schnorr** — `27000`

## Changes

- **`core`** — Replace `TURETZKY_UPPER_GAS_LIMIT` with `TURETZKY_UPPER_GAS_LIMIT_BLS` / `TURETZKY_UPPER_GAS_LIMIT_SCHNORR` and a `SignatureType` enum (`turetzky_upper_gas_limit()`, `total_gas_estimate()`, `label()`, `ALL`). Adds unit tests.
- **`cli`** — Both backends (evmsketch and `--anvil`) now print a base estimate (state-change cost, no floor) followed by a per-scheme block with the total estimate and savings.
- **`anvil`** — `GasKillerReport` gains a shared `gaskiller_base_gas_estimate` plus `_bls`/`_schnorr` variants of estimate, cost, and savings, derived via a new `SignatureGasEstimate::compute` helper. Receipt filters now compare against the smaller (Schnorr) floor so any tx that could save under either scheme is retained.

## Notes

- **CSV schema change:** `GasKillerReport` columns changed (`gaskiller_gas_estimate` → `gaskiller_gas_estimate_bls` / `_schnorr`, etc.). Downstream CSV consumers need the new headers.
- **WASM untouched:** its `gas_estimate` deliberately excludes the floor (the frontend applies it), so it is unaffected.

## Testing

- `cargo check` / `cargo clippy -- -D warnings` clean on default members and the anvil-feature paths (anvil crate + CLI `--features anvil`).
- `cargo fmt --all --check` clean; new core unit tests pass.

### DIFF
--- a/crates/anvil/src/lib.rs
+++ b/crates/anvil/src/lib.rs
@@ -191,13 +191,8 @@ impl SignatureGasEstimate {
         gas_used: u64,
         approx_gas_price_per_unit: f64,
     ) -> Self {
-        let gas_estimate = signature_type.total_gas_estimate(base_estimate);
-        let gas_savings = gas_used.saturating_sub(gas_estimate);
-        let percent_savings = if gas_used > 0 {
-            (gas_savings * 100) as f64 / gas_used as f64
-        } else {
-            0.0
-        };
+        let (gas_estimate, gas_savings, percent_savings) =
+            signature_type.savings(base_estimate, gas_used);
         Self {
             gas_estimate,
             estimated_gas_cost: approx_gas_price_per_unit * gas_estimate as f64,
--- a/crates/anvil/src/lib.rs
+++ b/crates/anvil/src/lib.rs
@@ -52,8 +52,8 @@ fn impl_slot() -> U256 {
 }
 
 use gas_analyzer_core::{
-    Opcode, RevertingContext, StateUpdate, TURETZKY_UPPER_GAS_LIMIT, compute_state_updates,
-    encode_state_updates_to_abi, encode_state_updates_to_sol,
+    Opcode, RevertingContext, SignatureType, StateUpdate, TURETZKY_UPPER_GAS_LIMIT_SCHNORR,
+    compute_state_updates, encode_state_updates_to_abi, encode_state_updates_to_sol,
 };
 use gas_analyzer_rpc::get_tx_trace;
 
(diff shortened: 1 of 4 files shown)


BreadrichEngels (approved):
Clean, well-factored feature — the `SignatureType` abstraction centralizes the base+floor arithmetic (`total_gas_estimate`), both CLI backends now compute a floor-free base and add the per-scheme floor at report time, and the core unit tests (including the compile-time `const { assert!(SCHNORR < BLS) }`) lock the invariants. LGTM.

Things I checked that hold up:
- **BLS is a faithful superset of the old behavior.** `TURETZKY_UPPER_GAS_LIMIT_BLS = 250000` equals the old single constant, so every `_bls` figure matches what the report produced before — this only *adds* the Schnorr view rather than changing existing numbers.
- **The receipt-filter change is correct.** Dropping the block/tx pre-filter from the 250k floor to the smaller Schnorr 27k floor is the right direction: it retains any tx that could show savings under *either* scheme, and `saturating_sub` cleanly yields 0 savings for the BLS case where `base + 250000 > gas_used`. (Side effect: block mode now analyzes the 27k–250k-gas band it previously skipped — intended, just slightly more work per block.)
- **No downstream breakage from the const rename.** I searched gas-killer/service for `TURETZKY_UPPER_GAS_LIMIT` — zero hits — and all four in-repo references (core re-export, cli, encoding, anvil) are updated here. Combined with the WASM/evmsketch `gas_estimate` deliberately excluding the floor, the service's evmsketch integration is unaffected, so no lockstep bump is needed. Nice that this shared-crate rename stays contained.

**Notes (non-blocking):**
- **CSV schema break** (as you flag): `gaskiller_gas_estimate` → `_bls`/`_schnorr` (+ `_base`), same for cost/savings/percent. Anything parsing the anvil reporter's CSV — dashboards, analysis scripts — needs the new headers. Worth a heads-up wherever those CSVs are consumed.
- **Percent-savings computed two ways.** `SignatureGasEstimate::compute` uses `(gas_savings * 100) as f64 / gas_used as f64` (integer-mul first) while the CLI evmsketch loop uses `(gas_savings as f64 / gas_used as f64) * 100.0` (float first). Identical for realistic gas values (the integer form only risks u64 overflow above ~1.8e17 gas, impossible on-chain) — purely a consistency nit if you feel like unifying.

Bench bot reports no regressions (all <7%, under the 15% gate). CI checks aren't reporting on the head commit yet — confirm `fmt`/`clippy -D warnings`/tests are green on both the default and `--features anvil` paths before merge. Approving.

rubydusa (approved):

PR #165, opened 2026-07-05
Prestate fast path: fall back on reverted frames + anvil-backed integration tests

RonTuretzky:
Finishes #163 (stacked on `hudsonhrh/prestate-hybrid-extraction`).

## Soundness fix: reverted frames must force the struct-log fallback

`classify_prestate_eligibility` never looked at `CallFrame.error`. Probing a real node (anvil 1.7.1) shows why that breaks the "never emits an unsound diff" invariant:

- **Reverted top-level call**: `prestateTracer` diffMode returns an empty diff, but anvil does **not** prune the reverted root frame's logs from `callTracer` `withLog` output. The classifier saw "no cross-contract storage, no disqualifying frames" → Eligible → the fast path emitted the reverted call's LOG on top of an empty storage diff — **an event that never happened**.
- **Caught-revert DELEGATECALL child**: the child's writes roll back (absent from the net diff), but `compute_state_updates` is revert-unaware and still extracts the child's SSTOREs/LOGs at target depth — so an Eligible classification silently diverged from the struct-log path.

The classifier now returns `Fallback` when the root frame or any transparent (`DELEGATECALL`/`CALLCODE`) target-depth frame carries `error`/`revertReason`. Reverted `STATICCALL`s stay eligible (they can't affect state on either path). Fallback = the exact previous behaviour, so parity is preserved by construction.

Also fixed the stale doc rationale for CREATE/CREATE2 (the `Create`/`Create2` StateUpdate variants exist since #155; fallback is still required because a net diff can't reconstruct the replayable initcode op) and documented the undetectable-`TSTORE`-warning caveat.

## Latency: cheap tracers now fetched concurrently

`try_prestate_fast_path` awaited the two independent `debug_traceCall`s sequentially; `tokio::try_join!` cuts the fast path to one round-trip of added latency, matching the concurrency pattern the surrounding code already uses.

## Tests: the layers that previously only type-checked

#163's core logic had 15 unit tests, but the `rpc` helpers and the `evmsketch` dispatcher had zero in-repo coverage (E2E validation lived in an external harness). Added:

- **4 core unit tests** for the new revert classification (root error, revertReason-only, reverted delegatecall child, reverted staticcall still eligible).
- **6 anvil-backed integration tests** in `evmsketch` that spawn a local anvil (CI already installs foundry via `foundry-rs/foundry-toolchain`, so they run un-ignored), deploy hand-assembled runtime bytecode via `anvil_setCode`, and run **both real extraction paths** against the same call:
  - eligible call → fast path output is the net-deduplicated, slot-sorted equivalent of the struct-log output (zeroing kept, net-zero write omitted, intermediate values collapsed)
  - delegatecall log ordering → `position`/`index` interleaving reproduces true emission order (root, lib, root) against a real tracer, matching struct-log
  - regular CALL / reverted call / caught-revert delegatecall → classifier forces fallback and hybrid output is byte-identical (ABI-encoded) to the struct-log path; **both revert tests fail without the classifier fix** (verified by temporarily removing it)
  - rpc helper shape contract → diff carries changed slots, frame carries logs with `position` populated

## Verification

- `cargo test` (default workspace): 22 core + 9 estimator + 20 evmsketch, all green
- `cargo test -p gas-killer-wasm --lib`: 28 green; `cargo clippy` (workspace + wasm) `-D warnings` clean; `cargo fmt --check` clean
- The `test-native`/`test-wasm` failures on #163 are the Alchemy monthly-quota 429 hitting the pre-existing `RPC_URL` integration tests — unrelated to this code. Long-term fix is #109 (local anvil instead of live RPC); the new tests here follow that direction.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RonTuretzky:
@hudsonhrh — this PR stacks on top of your `hudsonhrh/prestate-hybrid-extraction` branch (#163) to close out the two gaps I found while reviewing it. Merging this into your branch first will update #163 in place. Summary of the three commits:

### 1. Soundness fix — reverted frames must force the struct-log fallback (`36cbb1c`)
`classify_prestate_eligibility` never inspected `CallFrame.error`, which breaks the "never emits an unsound diff" invariant. Verified empirically against anvil 1.7.1 with hand-assembled bytecode:
- **Reverted top-level call**: `prestateTracer` diffMode returns an empty diff, but anvil does **not** prune the reverted root frame's logs from `callTracer` output. The old classifier returned `Eligible` → the fast path emitted a LOG for an event that was rolled back, on top of an empty storage diff.
- **Caught-revert DELEGATECALL child**: the child's writes vanish from the net diff, but `compute_state_updates` is revert-unaware and still extracts them at target depth — so `Eligible` silently diverged from the struct-log path.

Now any error/revertReason on the root or a transparent DELEGATECALL/CALLCODE frame forces `Fallback`. A reverted **STATICCALL** stays eligible (read-only, can't affect state on either path). Fallback = the exact prior behaviour, so parity holds by construction. Also corrected the stale CREATE/CREATE2 rationale (the `Create`/`Create2` variants exist since #155; fallback is still needed because a net diff can't reconstruct the replayable initcode op) and documented the undetectable-`TSTORE`-warning caveat.

### 2. Concurrency + the tests the rpc/dispatcher layers were missing (`3f655e5`)
- `try_prestate_fast_path` awaited the two independent `debug_traceCall`s sequentially; `tokio::try_join!` cuts the fast path to one round-trip of added latency.
- Your core logic had 15 unit tests, but the `rpc` helpers and the `evmsketch` dispatcher only type-checked (E2E lived in the external harness). Added **4 core unit tests** for the new revert classification and **6 anvil-backed integration tests** that spawn a local anvil (CI installs foundry, so they run un-ignored), deploy runtime bytecode via `anvil_setCode`, and run **both real extraction paths** against the same call:
  - eligible call → fast-path output is the net-deduplicated, slot-sorted equivalent of the struct-log output (zeroing kept, net-zero write omitted)
  - delegatecall log ordering → `position`/`index` interleaving reproduces true emission order (root, lib, root) against a real tracer, matching struct-log
  - regular CALL / reverted call / caught-revert delegatecall → classifier forces fallback and hybrid output is byte-identical (ABI-encoded) to struct-log; **both revert tests fail without the classifier fix** (confirmed by removing it)
  - rpc helper shape contract → diff carries changed slots, frame carries logs with `position` populated

### 3. Copilot review follow-ups (`5efbb96`)
- `LocalAnvil::spawn` now awaits `tokio::time::sleep` during the readiness poll instead of `std::thread::sleep` (yields rather than blocking the executor thread) and captures the last `get_chain_id` error in the panic, since anvil's stdout/stderr are suppressed.
- Tightened the dispatcher doc to match the actual rule (reverted STATICCALL stays eligible).

### Verification
`cargo test` (default workspace) green; `cargo test -p gas-killer-wasm --lib` green; `cargo clippy` (workspace + wasm) `-D warnings` and `cargo fmt --check` clean. CI on this PR: lint / test-native / test-wasm all pass.

Note: #163's earlier red `test-native`/`test-wasm` were the Alchemy monthly-quota 429 on the pre-existing live-RPC tests, not your code — the quota reset with the new month and I re-ran those jobs, so #163 is green again now too. The new tests here follow the #109 direction (local anvil instead of live RPC).

bagelface:
Synced this with `main` now that #163 has merged (merge commit, not a rebase — didn't want to force-push over your branch). It's `MERGEABLE` again and the full suite including all 6 anvil tests is green locally. Heads-up on what changed underneath you, and on one design call I made differently that you should push back on if you disagree.

## What landed in #163 that overlaps this PR

#163 grew a fair bit after review. Three of your four changes are now upstream:

- **The revert guard** — same conclusion as yours, arrived at independently.
- **`tokio::try_join!`** on the two tracers — identical.
- **The CREATE/CREATE2 doc** — you were right that "no `StateUpdate` variant represents them" was wrong; `Create`/`Create2` have existed since #155. Your wording went in.

Your `revertReason` fallback was the better version of the guard and went in too — `error.as_deref().or(revert_reason.as_deref())`, now behind a `frame_failure` helper, and applied to the log-collection path as well, which had the same `error`-only blind spot. Your `reverted_staticcall_child_still_eligible` test is upstream as well; the other three had equivalents already.

So this PR is now essentially **the anvil test suite**, which is the part that was hardest to write and the part #163 most needed. The bytecode fixtures survived untouched.

## The design call I made differently

The bigger change is that the prestate path is no longer an optimisation applied underneath the other encodings — it's `StateEncoding::PrestateNet`, a third *signed representation* selected fleet-wide, because the encoded updates are hashed into the task digest ([review thread](https://github.com/gas-killer/gas-analyzer/pull/163#pullrequestreview-4866558714)).

That invalidates the "parity with the struct-log path" rationale this PR is built on, and it changes one of your tests:

**Caught-revert DELEGATECALL.** You made it force a fallback so the output stays byte-identical to `Legacy`. I didn't, because `Legacy` is *wrong* there — it replays the rolled-back store and log, which its own docs call out — so matching it means being bug-compatible. `PrestateNet` keeps its net form for that call instead: the reverted child's writes are absent from the net diff by construction, and its logs are excluded when collecting. It's also the case that matters most in practice, since a heavy-compute call containing a caught revert would otherwise fall back and time out.

Rather than delete your test, I repurposed it, because your fixture is the best evidence we have that the `Legacy` bug is real and reachable. It now pins all three encodings against the same call:

```
net form   [Store(2, 0x22)]                                  ← correct
canonical  [Store(2, 0x22)]                                  ← agrees
legacy     [Store(1, 0x42), Log1(0xdd), Store(2, 0x22)]      ← replays rolled-back store + log
```

That last line is your original expected value, now labelled as the known unsoundness rather than the reference. It's pinned rather than fixed because `Legacy` is frozen — changing it is a digest migration for every deployment still on it (gas-analyzer#178).

Worth noting your fixture caught *my* mistake while I was writing that up: I had the slots backwards in the first draft, forgetting DELEGATECALL runs in the catcher's storage context, and the test failed until I read the bytecode properly.

## Other mechanical adaptations

- `extract_state_updates_hybrid` takes an `encoding` argument now; tests pass `StateEncoding::PrestateNet`.
- `extract_both` became `extract_all`, returning net / legacy / canonical. `PrestateNet` falls back to the **canonical** encoder, not legacy, so the fallback tests now compare against that — they were asserting against the wrong encoder after the merge.
- Renamed `test_fast_path_matches_struct_log_on_eligible_call` → `test_net_form_differs_from_struct_log_on_eligible_call`. The assertions were already pinning the difference; only the name claimed equivalence.

## Two things I left alone

Your `std::thread::sleep` → runtime-yield fix from the Copilot round is intact. And the tests run un-ignored, so they add anvil spawn time and a flakiness surface to default `cargo test` — they were fast and stable across several local runs, just flagging it as something to watch in CI.

One last thing worth your attention: I ran these on **anvil 1.5.1**, you probed on **1.7.1**, and everything agreed. That's mildly reassuring for the cross-client tracer-equivalence question raised as a prerequisite for enabling `PrestateNet` fleet-wide — your finding that anvil doesn't prune reverted frames' logs holds across both versions.

Happy to revert the caught-revert change if you think the conservative fallback is the safer call — you've spent more time against real tracer output than I have.

BreadrichEngels (approved):
## Review: reverted-frame fallback + anvil-backed integration tests

**Approve.** After the #163 merge this is effectively test-only (the revert guard, concurrent tracers, and CREATE doc fix all landed upstream) plus a `tokio` `time` feature — and it's the part #163 most needed: real end-to-end coverage of the `rpc` helpers and the dispatcher, which previously only type-checked. I spot-checked the hand-assembled bytecode and the expected values, and the tests pin exactly the three-way divergence I raised conceptually on #163, now against a live node.

### The tests do the thing that mattered
- **`eligible_code`** decodes correctly (`SSTORE(1,0x42)`, `MSTORE`+`LOG1(0xaa)`, `SSTORE(2,5)` then `SSTORE(2,0)` net-zero, `SSTORE(5,0)` zeroing a pre-seeded slot), and the expected net form `[Store(1,0x42), Store(5,0), Log1(0xaa,…)]` vs legacy `[Store(1), Log, Store(2,5), Store(2,0), Store(5,0)]` is right: net-dedup, net-zero omission, and slot-sort-then-logs — the exact `Legacy` divergence I described, now proven against anvil rather than argued.
- **`test_hybrid_falls_back_on_reverted_call`** is the load-bearing one. It captures the real anvil quirk — a reverted root frame's logs are *not* pruned from `callTracer` (unlike geth) — so without the classifier's failure check the net form would emit a `LOG` for an event that rolled back. That the test fails with the guard removed is what makes it a regression guard rather than decoration; validating it against real-tracer behavior is precisely what the synthetic unit tests couldn't.
- **`test_delegatecall_log_ordering_matches_between_paths`** + the `logs_only(net) == logs_only(legacy)` cross-check is the one I most wanted: it validates the `position`/`index` log-order reconstruction against a real tracer, which is the "third byte-identity surface" I flagged on #163. Event order is the observable-on-chain invariant, and pinning it across paths is the right assertion.
- The fallback tests correctly compare against the **canonical** encoder (PrestateNet's paired struct-log encoder), not legacy — that's the right reference after the encoding rework.

### On the caught-revert design call (you invited pushback)
Keeping the net form for a caught-revert *transparent* child — rather than the conservative fallback — is the correct call, and sound: the reverted child's writes are absent from the `prestateTracer` net diff by construction, and `collect_target_depth_logs` excludes an error-carrying frame's logs, so the net form is exactly the surviving effect (`[Store(2,0x22)]`), which is ground truth. It agrees with canonical and diverges only from `Legacy` — and `Legacy` is the one that's *wrong* there (it replays the rolled-back store + log), so matching it would be bug-compatibility. It's also the practically important case, since forcing a fallback would make any heavy-compute call containing a caught revert time out — defeating the whole point. Pinning the `Legacy` output as a labelled known-unsoundness (frozen, migration-only per gas-analyzer#178) rather than the reference is exactly the right framing. I checked the storage-context reasoning too (DELEGATECALL runs in the catcher's storage, so the rolled-back `SSTORE(1)` hit the catcher's own slot 1) — the fixture is correct.

### One thing to watch (non-blocking; you already flagged it)
The anvil tests run **un-ignored in default `cargo test`**, so they hard-require `anvil` on `PATH` — a contributor or a future CI job without foundry gets a *failure*, not a skip (the `.expect(...)` panics, helpfully, but it's still red). Combined with a per-test anvil spawn and the bind-`:0`-then-release-then-anvil-binds port-reuse window, that's a real flakiness/friction surface on a path everyone runs. Options: gate them behind `#[ignore]` (run in a dedicated CI step) or a feature flag, share one anvil across the suite, or make it a hard project invariant that every `cargo test` environment installs foundry. It works in this CI today; I'd just not want a `cargo test` on a laptop without foundry to fail mysteriously. Your call — the current setup is defensible if foundry-everywhere is guaranteed.

Nice touch that the fixtures caught a real bug mid-writeup (the reversed slots), and that anvil 1.5.1 vs 1.7.1 agreed — mildly reassuring for the cross-tracer-equivalence prerequisite I raised before `PrestateNet` can be enabled fleet-wide, though that's still its own gate. Solid closeout of the extraction work's test debt.

bagelface:
Took the non-blocking note — it turned out to be a real regression rather than a judgement call, so it's fixed in `d1ce4dd`.

You wrote that the current setup is "defensible if foundry-everywhere is guaranteed." It isn't:

- `default-members` includes `crates/evmsketch` but **not** `crates/anvil`, so a plain `cargo test` runs these tests.
- Before this PR, no default-member test needed foundry — `gas-analyzer-anvil` lives outside the default set and `gas-analyzer-cli` gates its anvil tests behind a feature.

So this PR would have been the thing that first made `cargo test` require foundry, and the failure mode is a panic rather than a skip. Fixed by matching the idiom `gas-analyzer-cli` already uses:

- New `anvil` feature on `gas-analyzer-evmsketch`, off by default.
- The integration tests move into a `#[cfg(feature = "anvil")] mod anvil_integration`.
- New CI step: `cargo test -p gas-analyzer-evmsketch --features anvil`.

`cargo test` is back to 17 evmsketch tests and no external tooling; `--features anvil` runs all 23.

One thing that fell out of it: the lint job would no longer have seen this code at all, since `#[cfg]`-gated modules aren't compiled without the feature — so a clippy warning in ~500 lines of test code would have gone uncaught forever. Added `--features gas-analyzer-evmsketch/anvil` to the clippy invocation so the `-D warnings` gate still covers it. Clean locally.

Left the per-test anvil spawn and the bind-then-release port window alone. Spawn is ~30 ms and the whole gated suite runs in 0.2 s, so pooling would trade real complexity — shared state across tests that currently get a clean chain each — for no measurable gain. Happy to revisit if it proves flaky in CI.

Thanks for the close read on the caught-revert reasoning and the storage-context check — good to have that independently confirmed rather than taken on my word.

Issue #66, opened 2025-09-17
Add evmsketch/anvil integration tests for block-specific transaction simulation


bagelface:
### Summary
We need to add integrations tests for the Gas Killer Analyzer (evmsketch/anvil layer) to verify behavior when analyzing transactions that depend on the correct block number for simulation. Tests should ensure the analyzer correctly handles simulations that require specific block numbers to function properly.

### Tasks

- [ ]  Create a new dummy smart contract in `gk-alloy-adapter` with a function that only works when simulated at the correct block number.
- [ ]  Deploy the contract on a testnet and add the resulting transaction hash to the constants file.
- [ ]  Add a test that passes when analyzing the transaction at the correct block number.
- [ ]  Add a test that fails when analyzing the transaction at an incorrect block number.

PR #166, opened 2026-07-07
feat: unbounded simulation mode (SimProfile::Unbounded) — break the block gas limit off-chain, keep the payload inside one transaction

RonTuretzky:
Stacked on #165, now merged; this branch is synced with `main`.

## What

Gas Killer's core trade is compute-off-chain / state-diff-on-chain — but tracked functions were still simulated under the real chain env, so the block gas limit (and post-Osaka, the EIP-7825 2^24 tx cap) bounded what could be killed. `SimProfile::Unbounded` removes the bound: tracked functions simulate under pinned protocol limits (`2^40` block+tx gas), in exchange for the extracted payload having to fit in a single on-chain transaction. Unbounded Solidity; on-chain state bounded by what a transaction can actually carry.

## How

- **core**: `SimProfile { Chain, Unbounded }`, pinned `UNBOUNDED_*` constants (operators, this analyzer, and the SP1 slashing guest must use bit-identical env or fraud proofs falsely slash honest operators — see `docs/UNBOUNDED_MODE.md`), `validate_unbounded_cost()`.
- **rpc**: `_with_profile` variants of the three `debug_traceCall` helpers; the profile sets `tx.gas` + `blockOverrides.gasLimit` unconditionally (pinned protocol env, not a hint). Node must have its cap lifted (`anvil --disable-block-gas-limit`, `geth --rpc.gascap=0`); a clamping node falls into the #165 revert-fallback and extraction still returns `Ok` — with an empty or partial payload, which is the sharper failure mode pinned below, not "fails safe".
- **evmsketch**: `call_to_encoded_state_updates_with_evmsketch_profiled()` threads the profile through both extraction paths and enforces the payload budget. Payload gas estimation deliberately stays on the real chain env (`verifyAndUpdate` lands in a real block).
- **guest consistency**: mechanism implemented in BreadchainCoop/sp1-contract-call`ron/unbounded-env-overrides` (companion PR) — overrides are bound into `chainConfigHash` so a proof under different limits can't verify.

## Tests

- core unit tests for the profile + the payload budget gate.
- anvil-backed integration tests (`--disable-block-gas-limit`): a ~40M-gas busy-loop consumer OOGs under `Chain` (forces the #165 revert fallback) and extracts exactly `[Store, Log1]` under `Unbounded`; a two-slot writer is **accepted** (it fits), and a 1,000-slot writer is rejected with `PayloadTooExpensive`.

---

## Updates since review

**Renamed `SimProfile::UnboundedV1` → `Unbounded`** (and `UNBOUNDED_V1_*` → `UNBOUNDED_*`), which is
what the code has carried since the sync with `main`; this description had been left stale. The
suffix carried no wire meaning — nothing serialises the variant name, and the SP1 guest binds the
*values* into `chainConfigHash`, not the identifier — and unsuffixed matches the convention
`StateEncoding::{Legacy, Canonical, PrestateNet}` already sets for equally consensus-critical
variants. The "these are pinned protocol constants, changing them is a lockstep rollout, never a
silent edit" warning now lives explicitly in the `sim_profile` module docs. (See the review thread —
there is a reasonable argument for keeping a version in the type; this is a deliberate call, not an
oversight.)

**Pinned what a clamped node actually produces**, per the review ask. Probing a real node rather than
reasoning about it turned up something sharper than "fails safe":

| scenario (clamping node, `Chain` profile) | result |
|---|---|
| OOG before any write | `Ok([])` — empty |
| write, then OOG | `Ok([Store(1, 0x42)])` — **partial** |

Extraction returns `Ok` in both cases, identically across all three encodings, so nothing downstream
can distinguish a clamped result from a real one — and the partial payload *passes* the gate, because
one store is far inside the payload budget. Both are now pinned by
`chain_profile_oog_after_a_write_yields_a_partial_payload` and an added assertion in
`test_unbounded_profile_extracts_beyond_block_gas_limit`, and `docs/UNBOUNDED_MODE.md` plus the
`apply_sim_profile` docs now state the failure mode plainly instead of "fails safe".

## Payload bound: priced, not counted

The gate no longer requires a single-slot payload. What has to stay bounded is the **gas needed to
apply the diff**, so a transition may write as many slots as fit under
`UNBOUNDED_PAYLOAD_GAS_BUDGET` — EIP-7825's per-transaction cap, `2^24`. Roughly **650 cold writes**
instead of one (a little fewer when slots and values are dense, since their bytes cost more to carry).

That makes the profile symmetric about EIP-7825: it *lifts* the cap for the off-chain simulation,
where it protects nothing, and *enforces* it on the payload, where it is binding. Using the protocol
limit rather than a tuned number also keeps the budget from becoming a knob.

The single-slot commitment pattern (solidity-sdk#51) stays available for contracts whose state is too
large for one transaction — but it is **no longer a requirement of this mode**, which matters given
that #51 is still an open PR.

**The bound is analytic on purpose.** `estimate_applied_payload_gas` is a pure function of the
payload, not the revm estimate the same call computes for reporting: accept/reject is part of what
operators must agree on, and a verdict derived from live-fetched state could put two honest operators
on opposite sides of the boundary and split quorum. For the same reason the gate carries its own
`UNBOUNDED_COLD_SSTORE_COST` (22,100 — cold SLOAD + zero→nonzero SSTORE) rather than reusing
`heuristic`'s tuned warm figure, which is tuned for user-facing savings numbers; a gate that
under-prices a write would admit payloads that do not fit. A compile-time assert pins the gate's
price above the heuristic's so the two cannot drift into each other.

`CREATE`/`CREATE2` remain rejected — not a cost question, a net diff simply cannot reconstruct
replayable initcode.

Tests: budget boundary (largest fitting payload accepted, one store past it rejected), signature
floor and external call gas counted, tracker slot priced but reported separately, plus two
anvil-backed cases — a two-slot writer now **accepted**, and a 1,000-slot writer rejected with
`PayloadTooExpensive`.

### The analytic model needed two more terms than it shipped with

Being independent of live state is right, but it means the model is a hand-written reimplementation
of something the measured path gets for free. `estimate_state_changes_gas` runs the real encoded
payload through revm, so revm charges intrinsic calldata gas and the handler's decode loop
automatically — nobody has to think about them. A hand-written model contains exactly the terms
someone remembered to write down, and this one priced **execution only**:

| at the old 746-store boundary | gas |
|---|---|
| what the gate computed | 16,757,600 |
| payload calldata it did not charge for | +871,528 |
| handler decode/dispatch it did not charge for | +1,273,156 |
| `UNBOUNDED_PAYLOAD_GAS_BUDGET` | 16,777,216 |

So the gate would accept payloads needing ~18.4M gas — a task nobody can settle post-Osaka. Not an
unsound diff and not a false slashing, and invisible except within a few percent of the ceiling,
which is exactly where an admission gate lives. Both terms are now priced off the real encoded
payload (`UNBOUNDED_APPLY_BASE_GAS`, `UNBOUNDED_APPLY_GAS_PER_PAYLOAD_BYTE`, and per-byte intrinsic
gas or the EIP-7623 floor, whichever binds), which is why capacity is ~650 rather than ~746.

The durable part is `analytic_bound_dominates_measured_apply_cost`: it uses the measured path as an
**oracle** for the analytic one, asserting the bound never falls below what revm actually charges
across store-, log-, and mixed-shaped payloads from 1 to 650 updates, sparse and dense. It passes the
signature floor as zero on purpose — a fixed 250,000 would mask a missing per-update term until
payloads passed ~200 updates. It caught the log-dispatch term while this fix was being written, and
it fails if either constant is dropped. Future dropped terms fail a test instead of loosening the
gate.

Caveat recorded in the constants' docs: the oracle is the `StateChangeHandlerGasEstimator` handler —
what this analyzer's own gas figures run through — not the production `verifyAndUpdate`. If that
handler's decode loop is materially more expensive, the per-byte constant needs re-measuring against
it.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -28,7 +28,6 @@ jobs:
             --exclude indexer-resolver \
             --exclude indexer-service \
             --exclude indexer-web \
-            --features gas-analyzer-evmsketch/anvil \
             --all-targets -- -D warnings
 
       # A single clippy run only ever sees one feature configuration, so anything behind a
@@ -82,8 +81,11 @@ jobs:
       - name: cargo test (anvil)
         run: cargo test -p gas-analyzer-anvil -p gas-analyzer-cli --features gas-analyzer-cli/anvil --no-default-features
 
+      # `#[ignore]`d so a `cargo test` without foundry skips rather than fails; foundry is installed
+      # above, so run them explicitly here. Name-filtered to this module so the RPC-backed ignored
+      # tests (which need RPC_URL) are not pulled in.
       - name: cargo test (evmsketch anvil integration)
-        run: cargo test -p gas-analyzer-evmsketch --features anvil
+        run: cargo test -p gas-analyzer-evmsketch anvil_integration -- --ignored
 
   test-wasm:
     runs-on: ubuntu-latest
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -8,6 +8,7 @@ pub mod constants;
 pub mod encoding;
 pub mod heuristic;
 pub mod prestate;
+pub mod sim_profile;
 pub mod trace;
 pub mod types;
 
@@ -23,6 +24,10 @@ pub use heuristic::{
 };
 pub use prestate::{
     PrestateEligibility, build_state_updates_from_prestate, classify_prestate_eligibility,
+};
+pub use sim_profile::{
+    SimProfile, UNBOUNDED_BLOCK_GAS_LIMIT, UNBOUNDED_TX_GAS_LIMIT, UnboundedShape,
+    UnboundedShapeViolation, validate_unbounded_shape,
 };
 pub use trace::{
     compute_state_updates, compute_state_updates_canonical, copy_memory, parse_trace_memory,
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -26,8 +26,9 @@ pub use prestate::{
     PrestateEligibility, build_state_updates_from_prestate, classify_prestate_eligibility,
 };
 pub use sim_profile::{
-    SimProfile, UNBOUNDED_BLOCK_GAS_LIMIT, UNBOUNDED_TX_GAS_LIMIT, UnboundedShape,
-    UnboundedShapeViolation, validate_unbounded_shape,
+    SimProfile, UNBOUNDED_BLOCK_GAS_LIMIT, UNBOUNDED_COLD_SSTORE_COST,
+    UNBOUNDED_PAYLOAD_GAS_BUDGET, UNBOUNDED_TX_GAS_LIMIT, UnboundedCost, UnboundedCostViolation,
+    estimate_applied_payload_gas, validate_unbounded_cost,
 };
 pub use trace::{
     compute_state_updates, compute_state_updates_canonical, copy_memory, parse_trace_memory,
--- a/crates/core/src/sim_profile.rs
+++ b/crates/core/src/sim_profile.rs
@@ -151,6 +151,45 @@ const LOG_BYTE: u64 = 8;
 /// Intrinsic cost of the `verifyAndUpdate` transaction that carries the payload.
 const TX_BASE: u64 = 21_000;
 
+/// Intrinsic gas per byte of payload calldata: 4 per zero byte, 16 per non-zero byte
+/// (EIP-2028 pricing, i.e. 4 gas per EIP-7623 *token*).
+const CALLDATA_ZERO_BYTE: u64 = 4;
+const CALLDATA_NONZERO_BYTE: u64 = 16;
+
+/// EIP-7623 floor price per calldata token, where a token is one zero byte or a quarter of a
+/// non-zero byte. A transaction pays the *larger* of its execution total and this floor, so
+/// [`estimate_applied_payload_gas`] has to evaluate both.
+const CALLDATA_FLOOR_TOKEN: u64 = 10;
+const CALLDATA_FLOOR_NONZERO_TOKENS: u64 = 4;
+
+/// Fixed cost of entering `verifyAndUpdate`'s apply loop: decoding the two outer arrays and
+/// setting up memory, independent of what they hold.
+///
+/// Together with [`UNBOUNDED_APPLY_GAS_PER_PAYLOAD_BYTE`] this covers the difference between an
+/// op's raw EVM price and what it costs to *dispatch* that op out of an ABI-encoded payload — a
+/// term the reporting estimators never needed, because revm charges it automatically when it
+/// executes real calldata.
+pub const UNBOUNDED_APPLY_BASE_GAS: u64 = 2_000;
+
+/// Cost of decoding and dispatching one byte of encoded payload inside `verifyAndUpdate`:
+/// `calldataload`ing it, copying it into memory, and walking the element it belongs to.
+///
+/// Charged per payload byte rather than per update because that is what the work is proportional
+/// to — an update's decode cost tracks the bytes it occupies, not its existence. Measured against
+/// the `StateChangeHandlerGasEstimator` handler across store- and log-shaped payloads at 7.9–12.9
+/// gas per byte (the top of that range being logs, which carry more envelope per byte), so 14
+/// dominates every observed shape with ≥8% margin.
+///
+/// `analytic_bound_dominates_measured_apply_cost` in `gas-analyzer-estimator` pins this and
+/// [`UNBOUNDED_APPLY_BASE_GAS`] against that measurement, so neither can silently drift below what
+/// applying a payload really costs.
+///
+/// Caveat worth knowing: the measurement is against the estimator handler, which is what this
+/// analyzer's own gas figures run through — not the production `verifyAndUpdate` in the
+/// solidity-sdk. If that handler's decode loop is materially more expensive, this constant needs
+/// re-measuring against it.
+pub const UNBOUNDED_APPLY_GAS_PER_PAYLOAD_BYTE: u64 = 14;
+
 /// Summary of a payload that passed [`validate_unbounded_cost`].
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct UnboundedCost {
@@ -165,8 +204,10 @@ pub struct UnboundedCost {
     pub calls: usize,
     /// Number of `Log0`–`Log4` ops.
     pub logs: usize,
-    /// Upper bound on the gas needed to apply this payload on-chain, including the transaction's
-    /// intrinsic cost and the signature-verification floor.
+    /// Upper bound on the gas needed to apply this payload on-chain: the transaction's intrinsic
+    /// cost, the signature-verification floor, each op's execution cost, the intrinsic gas for the
+    /// encoded payload's own bytes, and `verifyAndUpdate`'s decode/dispatch overhead. See
+    /// [`estimate_applied_payload_gas`].
     pub applied_gas_upper_bound: u64,
 }
 
@@ -232,6 +273,21 @@ impl std::error::Error for UnboundedCostViolation {}
 /// A pure function of its arguments, which is the point: it decides whether a payload is acceptable,
 /// so every operator must reach the same verdict from the same payload. Anything that consulted live
 /// chain state could put two honest operators on opposite sides of the boundary.
+///
+/// Three things are priced, and the payload has to carry all of them:
+///
+/// 1. **Execution** — the raw EVM cost of each op ([`UNBOUNDED_COLD_SSTORE_COST`], the `LOG*`
+///    schedule) plus `external_call_gas`.
+/// 2. **Transport** — intrinsic gas for the ABI-encoded payload's bytes. This scales with the
+///    number of updates, so unlike the reporting estimators (which get it for free from revm
+///    executing real calldata) the gate must charge it explicitly or it under-prices every
+///    multi-slot payload.
+/// 3. **Dispatch** — [`UNBOUNDED_APPLY_BASE_GAS`] plus
+///    [`UNBOUNDED_APPLY_GAS_PER_PAYLOAD_BYTE`] for decoding the payload and walking it inside
+///    `verifyAndUpdate`.
+///
+/// The return value is `max(execution + transport, EIP-7623 floor)`, mirroring how a transaction is
+/// actually priced: a payload whose bytes dominate its execution pays the floor instead.
 pub fn estimate_applied_payload_gas(
     updates: &[StateUpdate],
     external_call_gas: u64,
@@ -239,7 +295,8 @@ pub fn estimate_applied_payload_gas(
 ) -> u64 {
     let mut gas = TX_BASE
         .saturating_add(signature_floor)
-        .saturating_add(external_call_gas);
+        .saturating_add(external_call_gas)
+        .saturating_add(UNBOUNDED_APPLY_BASE_GAS);
     for update in updates {
         gas = gas.saturating_add(match update {
             StateUpdate::Store(_) => UNBOUNDED_COLD_SSTORE_COST,
@@ -254,7 +311,26 @@ pub fn estimate_applied_payload_gas(
             StateUpdate::Create(_) | StateUpdate::Create2(_) => 0,
         });
     }
-    gas
+
+    // Transport and dispatch, both priced off the real encoded payload rather than a per-op
+    // approximation: that keeps dense and sparse slot values honest, covers a log's data bytes,
+    // and stays correct if the encoding changes. The payload is what a `verifyAndUpdate`
+    // transaction carries, bar the selector and attestation, which ride inside `signature_floor`.
+    let payload = crate::encoding::encode_state_updates_to_abi(updates);
+    gas = gas.saturating_add(
+        (payload.len() as u64).saturating_mul(UNBOUNDED_APPLY_GAS_PER_PAYLOAD_BYTE),
+    );
+    let zero_bytes = payload.iter().filter(|byte| **byte == 0).count() as u64;
+    let nonzero_bytes = payload.len() as u64 - zero_bytes;
+    let transport = zero_bytes
+        .saturating_mul(CALLDATA_ZERO_BYTE)
+        .saturating_add(nonzero_bytes.saturating_mul(CALLDATA_NONZERO_BYTE));
+
+    let tokens =
+        zero_bytes.saturating_add(nonzero_bytes.saturating_mul(CALLDATA_FLOOR_NONZERO_TOKENS));
+    let floor = TX_BASE.saturating_add(tokens.saturating_mul(CALLDATA_FLOOR_TOKEN));
+
+    gas.saturating_add(transport).max(floor)
 }
 
 /// Enforce the `Unbounded` profile's payload invariant: applying the extracted updates on-chain must
@@ -408,9 +484,16 @@ mod tests {
     fn the_budget_boundary_is_where_arithmetic_says_it_is() {
         // Exactly at the cap passes; one more store does not. Pins the comparison as `>` rather
         // than `>=`, so a payload that precisely fills a transaction is still usable.
-        let per_store = UNBOUNDED_COLD_SSTORE_COST;
-        let room = UNBOUNDED_PAYLOAD_GAS_BUDGET - TX_BASE;
-        let n = (room / per_store) as usize;
+        //
+        // The boundary is found by search rather than division: per-store cost is no longer a
+        // single constant, since transport depends on how the encoded payload's bytes fall.
+        let mut n = 0usize;
+        while check(&(0..n + 1).map(|_| store()).collect::<Vec<_>>()).is_ok() {
+            n += 1;
+            assert!(n < 10_000, "boundary search must terminate");
+        }
+        assert!(n > 0, "at least one store must fit");
+
         let updates: Vec<StateUpdate> = (0..n).map(|_| store()).collect();
         let cost = check(&updates).expect("the largest payload that fits must be accepted");
         assert!(cost.applied_gas_upper_bound <= UNBOUNDED_PAYLOAD_GAS_BUDGET);
@@ -419,6 +502,38 @@ mod tests {
         assert!(
             check(&one_more).is_err(),
             "one store past the cap must fail"
+        );
+    }
+
+    #[test]
+    fn transport_and_dispatch_are_priced_on_top_of_execution() {
+        // The terms this gate exists to not forget. A single store's raw EVM cost is the cold
+        // SSTORE; what a transaction actually pays also includes the bytes that carry it and the
+        // loop that applies it, so the bound must sit strictly above the execution-only figure.
+        let updates = vec![store()];
+        let execution_only = TX_BASE + UNBOUNDED_COLD_SSTORE_COST;
+        let bound = estimate_applied_payload_gas(&updates, 0, 0);
+        assert!(
+            bound > execution_only + UNBOUNDED_APPLY_BASE_GAS,
+            "bound {bound} must exceed execution ({execution_only}) plus the fixed apply cost, \
+             leaving room for the payload's own bytes"
+        );
+
+        // Transport scales with payload bytes: a log carrying data costs more to carry than an
+        // empty one, beyond the 8 gas/byte the LOG opcode itself charges.
+        let empty_log = estimate_applied_payload_gas(&[log1()], 0, 0);
+        let fat_log = estimate_applied_payload_gas(
+            &[StateUpdate::Log1(IStateUpdateTypes::Log1 {
+                data: Bytes::from(vec![0xab; 256]),
+                topic1: B256::ZERO,
+            })],
+            0,
+            0,
+        );
+        assert!(
+            fat_log > empty_log + 256 * 8,
+            "256 data bytes must cost more than the LOG opcode's 8/byte alone: \
+             {fat_log} vs {empty_log}"
         );
     }
 
@@ -459,7 +574,17 @@ mod tests {
     fn empty_payload_is_valid() {
         let cost = check(&[]).unwrap();
         assert_eq!(cost.stores, 0);
-        assert_eq!(cost.applied_gas_upper_bound, TX_BASE);
+        // Not bare `TX_BASE`: an empty payload still encodes two array headers — 128 bytes, being
+        // two offsets and two zero lengths, of which only the offsets' low bytes are non-zero —
+        // and applying it still enters the decode loop. Both are costs a real transaction pays.
+        let transport = 126 * CALLDATA_ZERO_BYTE + 2 * CALLDATA_NONZERO_BYTE;
+        let dispatch = UNBOUNDED_APPLY_BASE_GAS + 128 * UNBOUNDED_APPLY_GAS_PER_PAYLOAD_BYTE;
+        assert_eq!(
+            cost.applied_gas_upper_bound,
+            TX_BASE + transport + dispatch,
+            "an empty payload costs the tx base plus the cost of carrying and decoding two \
+             empty array headers"
+        );
     }
 
     #[test]
--- a/crates/core/src/sim_profile.rs
+++ b/crates/core/src/sim_profile.rs
@@ -12,8 +12,8 @@
 //! needed to apply the diff, so a transition may write as many slots as fit
 //! under [`UNBOUNDED_PAYLOAD_GAS_BUDGET`]. Consumers whose state is too large
 //! for that can commit it into fewer slots — in the limit, the single-slot
-//! commitment pattern (solidity-sdk PR #51) — and fan out across contracts via
-//! multi-call forwarding (solidity-sdk PRs #47/#48), but neither is required.
+//! commitment pattern (solidity-sdk PR #51) — but that pattern is an option for
+//! contracts that need it, not a requirement of this mode.
 //!
 //! # Determinism is protocol-critical
 //!
@@ -160,8 +160,8 @@ pub struct UnboundedCost {
     pub tracker_stores: usize,
     /// Number of `Call` ops. These re-execute **on-chain at real gas prices**
     /// when the payload is applied — unbounded compute inside a `Call` is NOT
-    /// killed. Forwarded multi-call sub-payloads (`applyForwardedUpdates`)
-    /// ride in this category by design.
+    /// killed, and is charged against the budget at the gas the trace measured
+    /// for it.
     pub calls: usize,
     /// Number of `Log0`–`Log4` ops.
     pub logs: usize,
--- a/crates/core/src/sim_profile.rs
+++ b/crates/core/src/sim_profile.rs
@@ -4,12 +4,16 @@
 //! under. `Chain` mirrors the real chain (today's behaviour everywhere).
 //! `Unbounded` is the Gas Killer *unbounded execution* profile: the tracked
 //! function is simulated with gas limits far above any real block, so
-//! arbitrarily heavy Solidity can be executed off-chain — provided its **net
-//! effect** still fits the Gas Killer on-chain payload shape (at most one
-//! storage write per consumer contract, plus external calls and logs; see
-//! [`validate_unbounded_shape`]). The intended consumer shape is the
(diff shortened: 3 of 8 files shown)


BreadrichEngels (commented):
## Review: unbounded simulation mode (`SimProfile::Unbounded`)

Well-built, and the consensus-criticality is handled up front this time — the `sim_profile.rs` module doc lays out the "operators + analyzer + SP1 guest must use bit-identical env or honest operators get falsely slashed" invariant exactly right, which is the thing I'd otherwise have led with. I read the core, rpc, and evmsketch changes; the vendored `flat`/guest side and the full anvil test bodies I sampled rather than line-read. COMMENT rather than approve: this is the fraud-proof-determining simulation env, so it sits behind the same hard rollout + audit gate as the rest of the slashing path, and I have one fail-safe contract to pin down.

### Verified sound
- **Pinned constants with real reasoning.** `2^40` for both block and tx gas (not `u64::MAX`) so intrinsic-gas/refund/floor arithmetic — all `u64` in revm — can't overflow, while still admitting ~27 GB of EIP-7623-priced calldata. Deliberately ignoring the EIP-7825 `2^24` tx cap is correct (that cap protects block building, irrelevant to an off-chain sim). Good.
- **The shape gate is the load-bearing safety invariant and it's right.** `validate_unbounded_shape` enforces ≤1 non-tracker `Store` + no `CREATE`/`CREATE2`, with the `STATE_TRACKER_SLOT` correctly exempted (every `trackState` diff carries it, and `verifyAndUpdate`'s own modifier rewrites it idempotently). That's what makes "burn a terabyte of gas off-chain" sound: the thing that lands on-chain is provably O(1) SSTOREs + bounded calls/logs, and a non-commitment-shaped consumer fails loudly (`TooManyStores`) rather than signing a payload that costs O(slots) on-chain. `Call` ops passing is correct — they re-execute on-chain at real gas, so unbounded compute inside a `Call` is *not* killed, and forwarded multi-call sub-payloads ride there by design. Tests cover the tracker exemption, 0/1/2 stores, create/create2, empty. Enforcement runs in `..._profiled` *before* encoding/estimating.
- **Threading is clean and backward-compatible.** `apply_sim_profile` sets tx-gas on the request and block-gas in `blockOverrides` unconditionally under `Unbounded`; the three `_with_profile` rpc variants delegate to `Chain` from the old signatures; `..._mode` delegates to `..._profiled(Chain)`. So `Chain` is byte-for-byte unchanged, the profile threads through *both* extraction paths, and the gas estimate correctly stays on the real chain env (the payload lands in a real block).

### The one contract to pin: what a clamping node actually returns
The safety story is "a node without its cap lifted makes heavy calls OOG → revert → #165 fallback → never an unsound diff." I want to nail the *result* of that fallback, because the anvil test asserts the **classifier verdict** (`Fallback(reason contains "reverted")`) but I couldn't confirm it asserts the **extraction result**. Walking it: a clamping node OOGs the top-level call → prestate path sees an empty diff + errored root → `#165` forces the struct-log fallback → which for `PrestateNet`/`Canonical` is the *canonical* encoder. From #165's `test_hybrid_falls_back_on_reverted_call` (net == canonical for a reverting call), canonical appears to return an **empty** diff for a failed top-level call, not an `Err`.

If that's right, the claim holds in the letter — an empty payload isn't *unsound* — but the operational consequence is sharper than "fails safe": a clamping (misconfigured) node signs an **empty** payload while an honest node signs `[Store, Log]`, so the two **diverge**. A minority of clamping nodes is simply excluded from quorum (safe). But a *majority* clamping would form quorum on an empty no-op payload — silently dropping the transition's effect on-chain, which the shape gate can't catch (empty passes). So **"every operator runs a cap-lifted node" is a hard consensus requirement, not a performance nicety**, and needs enforcement/monitoring at deploy. Two asks: (1) have the OOG anvil test assert the *extraction result* under `Chain` (Err vs empty), so the fail-safe contract is pinned rather than inferred; (2) make the doc say plainly that the failure mode is a divergent/empty payload from a mis-provisioned node, not just "OOG."

### Standing gates (before `Unbounded` is enabled anywhere)
- **Merge is a no-op for production** — `Chain` is the default and nothing selects `Unbounded` until the service explicitly calls `..._profiled(Unbounded)`, same gating discipline as `PrestateNet`. Good.
- **SP1-guest lockstep is mandatory and atomic.** The sim env now determines both the signed digest *and* the fraud oracle, so the companion `sp1-contract-call` PR (`chainConfigHash` binding the overrides — the right mechanism) must land bit-identical and roll out in lockstep with any operator enabling `Unbounded`. A guest on the real header's `gas_limit` while operators run lifted limits = false-slashing honest operators, exactly as your doc says. This is the highest-stakes coupling here.
- **Audit as a unit** — this + the guest env-override + the shape gate should be in the same external-audit scope as the slasher (#66); the shape gate is the only thing standing between "unbounded off-chain" and "unbounded on-chain cost," so it and the guest's matching enforcement are load-bearing together.

### Minor
- **Naming vs versioning.** The PR calls it `UnboundedV1` / `UNBOUNDED_V1_*`, but the code is `SimProfile::Unbounded` / `UNBOUNDED_*` with no version in the type — while the doc stresses "changing the constants … must ship as a new version." Consider tagging the variant (`UnboundedV1`) so a future limit change is a distinct type the compiler and the digest can tell apart, mirroring how `chainConfigHash` versions it guest-side; today a constant edit would be a silent reinterpretation of the same variant.

Genuinely strong work on the determinism framing — that was the hard part and it's front-and-center. The gate is entirely about the cross-repo/guest lockstep and the node-provisioning requirement, not the code in this diff.

BreadrichEngels (changes requested):
## Re-review — `SimProfile::Unbounded`, now with the priced payload gate

The two asks from my prior review are addressed well, and the shape→priced redesign is the right direction. But the redesign introduced one concrete correctness gap in the load-bearing gate itself, so I'm requesting changes on that single point. Everything else below is either verified-sound or acknowledgement.

### My earlier asks — resolved

- **Clamped-node behaviour is now pinned, and you found something sharper than I did.** Probing a real node surfaced the *partial* payload (`Ok([Store(1, 0x42)])` after a mid-call OOG), not just the empty case I'd reasoned to — and a partial payload *passes* the gate, so it's genuinely indistinguishable downstream. `chain_profile_oog_after_a_write_yields_a_partial_payload` + the added assertion pin the extraction result rather than just the classifier verdict, which is exactly what I wanted. `docs/UNBOUNDED_MODE.md` now states the divergent/partial failure mode plainly. Good.
- **Naming.** You went the opposite way from my nit (dropped the suffix rather than adding it), and the argument is sound: nothing serialises the variant name, the guest binds the *values* into `chainConfigHash`, and `Unbounded` matches the `StateEncoding::{Legacy,Canonical,PrestateNet}` convention. With the "pinned protocol constants, lockstep or nothing" warning now explicit in the module docs, I'm satisfied — deliberate call, not an oversight.

### Determinism of the new gate — verified sound

`estimate_applied_payload_gas` is a pure function of `(updates, external_call_gas, signature_floor)`: the payload and the trace-measured call gas are both fixed by the pinned sim env, and the floor is a scheme constant. No live-state read, so two honest operators can't land on opposite sides of the boundary. The compile-time `assert!(UNBOUNDED_COLD_SSTORE_COST > heuristic::WARM_SSTORE_COST)` pinning the gate's price above the reporting heuristic is the right guard against the two drifting into each other. SSTORE priced at the cold zero→nonzero max (22,100) and LOG priced at protocol rates are both conservative. This is the property that matters most and it holds.

### Blocking: the upper bound omits the payload's calldata gas

`UnboundedCost.applied_gas_upper_bound` is documented as "the gas needed to apply this payload on-chain, including the transaction's intrinsic cost," and the gate's whole contract (per the module docs and the call-site comment — *"signing a payload that cannot be mined produces a task nobody can settle"*) is that a payload it accepts fits under EIP-7825's `2^24` per-transaction cap. But `estimate_applied_payload_gas` only adds `TX_BASE` (21,000) for intrinsic — and Ethereum intrinsic gas is `21,000 + per-byte calldata cost`. The payload rides on-chain **as calldata** to `verifyAndUpdate`, and that calldata cost is part of the same `2^24` the gate checks against. It isn't counted.

Because SSTOREs are charged at exactly the cold cost with no headroom for the realistic all-cold commitment case, the omission is reachable near the boundary:

- Budget `2^24 = 16,777,216`. Max stores accepted ≈ `(16,777,216 − 21,000 − sig_floor) / 22,100 ≈ 750`.
- Each `Store` is `(bytes32 slot, bytes32 value)` = 64 bytes of ABI content (more with array/enum framing). At standard calldata pricing (16 gas/nonzero byte) that's ≥ ~1,024 gas/store of calldata the estimate never adds.
- So a ~750-store payload is estimated at ~16.7M (**accepted**) but its real transaction is ~16.7M + ~0.8–1.5M calldata ≈ **17.5–18.3M > 2^24** — i.e. it cannot set a valid `gas_limit` and **cannot be mined**. The gate admits exactly the payload it exists to reject.

The band is narrow (roughly the top ~700→750 of the accepted store count; 500 genuinely fits, 1,000 is correctly rejected either way), and to be clear about blast radius: this is **deterministic**, so it does *not* split quorum, and the SP1 guest re-executes under the same limits so it does *not* falsely slash — the failure is liveness: a quorum can sign a transition whose settlement tx is unmineable, silently stranding that transition. Still, that's precisely the outcome the gate's docstring promises to prevent, so I'd like it closed before this can be enabled.

**Suggested fix (deterministic, small):** you already have `encode_state_updates_to_abi` at the call site — price the encoded payload's calldata and add it to the bound, e.g. charge every encoded byte at the nonzero rate (16, or the EIP-7623 floor equivalent) as a conservative upper bound. Encoded length is a pure function of the payload, so this keeps the verdict operator-agnostic. A boundary test asserting a max-accepted payload's *real* apply cost (incl. calldata) is `≤ 2^24` would pin it — today `the_budget_boundary_is_where_arithmetic_says_it_is` only checks the estimator against itself, which is why the gap slipped through.

### Smaller notes (non-blocking)

- **`signature_floor` is hard-wired to `SignatureType::Bls.turetzky_upper_gas_limit()`** at the call site, while the payload is signed independently of scheme. The docstring says to pass "the larger scheme's when the scheme is not yet known" — please confirm BLS is in fact the larger floor of the two (BLS pairing verification usually is); if Schnorr's upper floor can exceed it, this under-prices the signature component. A `max(bls, schnorr)` would make the intent self-evident and drift-proof.
- **`external_call_gas` is trace-measured under the *Unbounded* env.** For a `Call` whose sub-execution branches on `gasleft()`, the gas measured under lifted limits could differ from what it costs on-chain. Almost certainly a non-issue (opcode costs are env-independent), but since it feeds the gate, worth a line in the docs noting the assumption that a Call's consumed gas is env-invariant.

### Standing gates — unchanged from prior review

Still behind the same hard rollout + audit gate: (1) merge is a no-op for production (`Chain` remains default; nothing selects `Unbounded` until the service opts in); (2) the SP1-guest `chainConfigHash` override binding must land bit-identical and roll out fleet-wide in lockstep — the sim env determines both the signed digest and the fraud oracle; (3) this + the guest env-override + the payload gate should be one external-audit unit with the slasher (#66). The gate is the only thing between "unbounded off-chain" and "unbounded on-chain cost," which is exactly why the calldata gap above matters.

Strong iteration overall — the determinism framing and the clamped-node diligence are excellent. Fix the calldata term in the bound and I'll flip to approve (subject to the standing cross-repo/audit gates, as before).

bagelface:
@BreadrichEngels — two things from your review need your attention again, and one of them
invalidates the part you signed off on. Also note the branch was **rebased onto `main`**, so the SHAs
your review is anchored to are gone; current history is `7a2c6e8..b6d2d50`.

## 1. Naming: we're keeping `Unbounded`

Your point stands on its merits — a version in the type would let the compiler and the digest tell
two limit sets apart, and "constant edit = silent reinterpretation" is a real failure mode. We're
going the other way anyway, deliberately:

- Nothing serialises the variant name. The guest binds the *values* into `chainConfigHash`
  (`ChainConfigWithEnvOverrides`), so a proof produced under different limits fails verification
  rather than diverging quietly. A half-rolled fleet doesn't produce mismatched-but-valid proofs.
- `StateEncoding::{Legacy, Canonical, PrestateNet}` already sets the unsuffixed convention for
  equally consensus-critical variants.
- The "pinned protocol constants, changing them is a lockstep rollout, never a silent edit" warning
  is explicit in the `sim_profile` module docs.

`docs/UNBOUNDED_MODE.md` no longer prescribes `UnboundedV2` — that section described a design the
code didn't have, which is partly how the description drifted from the code in the first place. Worth
knowing: the rename was never a real commit. It rode inside merge `56beb30` as an evil merge, with an
empty marker commit alongside it. That's now folded into `7a2c6e8`, which introduces the final name
directly.

If you still want the version in the type, say so and we'll take it as a follow-up rather than
re-open it here.

## 2. The shape gate you verified no longer exists — please re-read

This is the ask. Your strongest paragraph was *"the shape gate is the load-bearing safety invariant
and it's right"* — ≤1 non-tracker `Store`, `STATE_TRACKER_SLOT` exempted, `TooManyStores` on
violation. `941f00a` replaced it with a **priced** gate: any number of writes, as long as applying
the diff costs no more than `UNBOUNDED_PAYLOAD_GAS_BUDGET` (EIP-7825's `2^24`). `validate_unbounded_shape` →
`validate_unbounded_cost`; `TooManyStores` → `PayloadTooExpensive`.

The reasoning: it makes the profile symmetric about EIP-7825 (lift the cap where it protects nothing,
enforce it where it binds), uses the protocol's own limit rather than a tuned number, and de-couples
the mode from solidity-sdk#51, which is still open. But your soundness analysis was of a different
invariant, so it needs fresh eyes rather than carry-over.

**One thing to be skeptical of, because we already were.** The gate must be analytic — a verdict from
live-fetched state could put two honest operators on opposite sides of the boundary and split quorum
— which makes it a hand-written reimplementation of a cost model the measured path gets for free.
`estimate_state_changes_gas` runs the real payload through revm, so revm charges intrinsic calldata
and the handler's decode loop automatically. A hand-written model contains exactly the terms someone
wrote down, and the first version priced execution only:

| at the 746-store boundary as first shipped | gas |
|---|---|
| what the gate computed | 16,757,600 |
| payload calldata, uncharged | +871,528 |
| handler decode/dispatch, uncharged | +1,273,156 |
| `UNBOUNDED_PAYLOAD_GAS_BUDGET` | 16,777,216 |

It would have accepted payloads needing ~18.4M gas — a task nobody can settle post-Osaka. Not an
unsound diff and not a false slashing, and invisible outside the top few percent of the range, which
is exactly where an admission gate lives. `b6d2d50` prices both terms off the real encoded payload,
which is why capacity is ~650 writes rather than ~746.

The durable fix is `analytic_bound_dominates_measured_apply_cost` (`crates/gas-estimator/src/lib.rs`):
it uses the measured path as an **oracle** for the analytic one, asserting the bound never falls
below what revm actually charges, across store-, log-, and mixed-shaped payloads from 1 to 650
updates, sparse and dense values. Signature floor passed as zero on purpose — a fixed 250,000 masks
a missing per-update term until ~200 updates, i.e. the broken model would have passed. It caught a
second missing term (log dispatch) while the fix was being written, and it fails if either constant
is dropped.

**Two judgment calls worth your scrutiny specifically:**

1. `UNBOUNDED_APPLY_BASE_GAS` (2,000) and `UNBOUNDED_APPLY_GAS_PER_PAYLOAD_BYTE` (14) are
   **calibrated, not derived** — the thing the priced-gate design was trying to avoid. Measured at
   7.9–12.9 gas/byte across shapes; 14 dominates with ≥8% margin. The property test converts them
   from tuned to verified, which seems like the right trade, but it is a trade.
2. The oracle is `StateChangeHandlerGasEstimator` — what this analyzer's own gas figures run through
   — **not** the production `verifyAndUpdate`. If the SDK's decode loop is materially more expensive,
   the per-byte constant needs re-measuring against it. Recorded in the constants' docs; it's the one
   thing this test cannot tell us.

## Your two explicit asks: done

- **Assert the extraction *result* under `Chain`, not just the classifier verdict.** Done in
  `cd6ed1d`, and probing a real node turned up something sharper than "fails safe": a clamping node
  that writes *then* OOGs returns `Ok([Store(1, 0x42)])` — a **partial** commitment, `Ok`, non-empty,
  indistinguishable from a correct result, and it passes the budget gate too. Pinned by
  `chain_profile_oog_after_a_write_yields_a_partial_payload`, with the `Ok([])` empty case pinned in
  `test_unbounded_profile_extracts_beyond_block_gas_limit`.
- **Say plainly that the failure mode is a divergent/empty payload.** Done, in
  `docs/UNBOUNDED_MODE.md` § Node requirement and the `apply_sim_profile` docs. The "fails safe"
  phrasing is gone from the PR description too.

## Standing gates, unchanged

Guest lockstep (sp1-contract-call#12, still open), operator node-provisioning enforcement, and audit
scoping alongside #66 are all being handled outside this PR — flagging so they don't read as
unresolved here. Merge remains a no-op for production: `Chain` is still the default and nothing
selects `Unbounded` outside tests.

BreadrichEngels (approved):
## Final re-review — blocker resolved, approving

My `REQUEST_CHANGES` blocker is fixed, thoroughly. I verified the fix against source at `b6d2d50` (rebased history `7a2c6e8..b6d2d50`), not just the write-up. Approving the analyzer code + gate; the rollout gates below are unchanged and remain preconditions to *enabling* `Unbounded`, not to merging this.

### The calldata gap is closed, and you found a second term

`estimate_applied_payload_gas` now returns `max(TX_BASE + standard_calldata + execution + extras, TX_BASE + EIP-7623_floor)`, where `extras` = `signature_floor + external_call_gas + UNBOUNDED_APPLY_BASE_GAS + 14·payload_bytes`, and transport is priced off the *real* `encode_state_updates_to_abi(updates)` bytes (4/zero, 16/non-zero). I worked the algebra against the actual EIP-7623 formula `TX_BASE + max(standard_calldata + execution, floor)`:

- **execution-binding branch:** code's non-floor term = `TX_BASE + standard_calldata + execution + extras ≥` real, since `extras ≥ 0` and each op is priced at its worst case (cold SSTORE 22,100, full LOG schedule).
- **floor-binding branch:** code's floor term = `TX_BASE + 10·tokens` = the real floor exactly, and the `.max` can only return `≥` it.

So the result dominates the true tx cost in both regimes — it's a real upper bound now. And the two-term boundary table you posted (`+871,528` transport, `+1,273,156` dispatch pushing a "746-store" payload to ~18.4M > 2²⁴) matches my own estimate and then goes further: I only flagged transport; the handler decode/dispatch term was yours to catch.

### The oracle test is the right durable guard

`analytic_bound_dominates_measured_apply_cost` using `estimate_state_changes_gas` (revm on the real payload) as the oracle for the analytic bound is exactly how I'd want this pinned — it makes "did we forget a term" a CI failure rather than an over-admitting gate. Two design choices I want to call out as correct:

- **`signature_floor = 0`** so a fixed 250k can't mask a missing per-update term until ~200 updates — that's the subtle one, and it's right (it's what surfaced the log-dispatch term).
- **Sweeping dense values and fat logs** covers the worst-case zero/non-zero split (transport) and the highest envelope-per-byte (dispatch), and it runs out to n=650, i.e. the entire admissible range, so the calibrated constants are validated everywhere they can bind.

### On your two flagged judgment calls

1. **Calibrated `UNBOUNDED_APPLY_BASE_GAS` / `…_PER_PAYLOAD_BYTE` (14, vs 7.9–12.9 measured).** Acceptable. Calibration is unavoidable for an analytic gate, and the property test converts these from "tuned" to "verified across the admissible range," which is the right trade. The ≥8% margin is the cushion; the test is the ratchet.
2. **The oracle is `StateChangeHandlerGasEstimator`, not production `verifyAndUpdate`.** This is the one real residual, and it's the same class of cross-boundary concern I've flagged from the start — the gate calibrates to (and prices the bytes of) this analyzer's own apply model, while the payload actually lands in the SDK's `verifyAndUpdate`. If that decode loop is materially >8% more expensive per byte, the cushion erodes near the ceiling. It's documented in the constant's doc, and it doesn't touch determinism or the guest (so no quorum split, no false-slash) — worst case is a near-boundary liveness stranding. **I'm making this an explicit item in the audit/cross-repo scope:** re-measure the per-byte dispatch constant against the real `verifyAndUpdate`, and ideally wire the oracle to the SDK artifact so the two can't drift without failing CI. That closes the last thing this test structurally can't.

### Standing gates — unchanged, still preconditions to enabling `Unbounded`

Same as my first review: (1) merge is a no-op for production — `Chain` is default and nothing selects `Unbounded` outside tests; (2) the SP1-guest `chainConfigHash` env-override binding (sp1-contract-call#12) must land bit-identical and roll out fleet-wide in lockstep; (3) audit this + the guest override + the payload gate as one unit alongside the slasher (#66) — with the `verifyAndUpdate` per-byte re-measurement from point 2 above folded in.

### Naming

Accepting your call to keep `Unbounded` — the reasoning (guest binds values into `chainConfigHash`, matches `StateEncoding::{…}`, warning is explicit in the module docs) is sound and I won't reopen it. If you want the version-in-type later, a tracking issue for the follow-up is enough; no need to block here.

Excellent iteration — you turned my one blocker into a fix plus a second catch plus a regression that makes the whole class of bug fail loudly from now on. Approving.

BreadrichEngels (approved):
## Re-review @ `b6d2d50` (rebased; prior anchors `e2e3b10`/`a0158678` are gone) — flipping to **approve**

My CHANGES_REQUESTED was gated on one thing: the priced bound omitted the payload's calldata gas, so it could admit a ~750-store payload whose real tx exceeds EIP-7825's `2^24` and can't be mined. I read the gate and the oracle test at this head rather than taking the summary on faith — it's fixed, and fixed the right way.

### The blocking calldata gap — closed, verified in code
`estimate_applied_payload_gas` now prices three things off the **real encoded payload** (`encode_state_updates_to_abi(updates)`), not a per-op approximation:
- **Transport** — EIP-2028 intrinsic calldata (`4`/zero, `16`/nonzero) — the term I flagged as missing.
- **Dispatch** — `UNBOUNDED_APPLY_BASE_GAS + 14·len` for the handler's decode/copy/walk — the *second* term you caught while fixing the first.
- Returns `max(exec + transport, EIP-7623 floor)`, with the floor modelled correctly (`tokens = zero + 4·nonzero`, `21000 + 10·tokens`). Transport (intrinsic, charged by the EVM at tx entry) and dispatch (in-handler `CALLDATACOPY`/decode) are genuinely distinct costs, so charging both is conservative, not double-counting. The `max` mirrors how a real tx is priced. This is right.

### The oracle test is the durable closure — and stronger than the boundary test I asked for
`analytic_bound_dominates_measured_apply_cost` asserts `analytic ≥ measured(revm)` across sparse/dense stores, fat logs, and mixed, at n = 1…650, with `signature_floor = 0` on purpose so no per-update term hides behind a fixed constant (a 250k floor would've masked a missing term until ~200 updates — exactly how the first model slipped through). Because the gate rejects `analytic > budget`, an accepted payload has `measured ≤ analytic ≤ 2²⁴` — i.e. it provably fits one transaction. That's precisely the property the docstring promises, now enforced rather than asserted. Nice touch using CANCUN so the `2²⁴` clamp doesn't OOG the large measurements. Determinism is intact (pure function of the payload; the compile-time `UNBOUNDED_COLD_SSTORE_COST > heuristic::WARM_SSTORE_COST` guard is still there).

### Naming — settled, keep `Unbounded`
You went the opposite way from my nit and the reasoning holds: nothing serialises the variant name, and the guest binds the *values* into `chainConfigHash`, so a half-rolled fleet fails verification instead of diverging quietly — a stronger guard than a type suffix, and consistent with `StateEncoding::{Legacy,Canonical,PrestateNet}`. No follow-up needed from me; don't re-open it.

### The one residual to carry into the enablement gate (not blocking this PR)
The per-byte/base constants (`14`/`2000`) are **calibrated, not derived** — acceptable, because the oracle test converts them from tuned to verified. But it validates them against `StateChangeHandlerGasEstimator` (this analyzer's handler), **not** the production `verifyAndUpdate` in solidity-sdk. If the real decode loop is materially more than 14 gas/byte, the gate under-prices and the failure mode is a quorum signing an unmineable settlement tx (liveness — deterministic, so no split quorum and no false slash, per your own analysis). It's documented in the constants' docs, and it's genuinely the one thing this test can't tell us. **Before `Unbounded` is enabled in production**, re-measure both constants against the real SDK handler — ideally by porting an equivalent `analytic ≥ measured` oracle into the solidity-sdk repo so it can't drift cross-repo. Folding this into the existing standing gates rather than holding the merge.

### Smaller notes (non-blocking)
- `signature_floor` is hardwired to `SignatureType::Bls.turetzky_upper_gas_limit()` — conservative, since BLS pairing verification structurally dominates Schnorr's single `ecrecover`, and the call-site comment now says so. A `max(bls, schnorr)` would be drift-proof if a third scheme is ever added, but fine as-is.
- Coverage boundary worth noting: the oracle passes `signature_floor = 0`, so the signature-verification + selector/attestation-calldata term rides entirely on `turetzky_upper_gas_limit()` being a true upper bound — it isn't cross-checked by the measured path here. Acceptable given the "upper" contract; just be aware the oracle validates the *payload-apply* terms, not the *signature* term.
- `external_call_gas` is trace-measured under the lifted env; opcode costs are env-invariant so this is almost certainly fine, but a `Call` that branches on `gasleft()` is the one exception — still worth a one-line doc note.

### Standing gates — unchanged, and I see they're tracked outside this PR
Guest `chainConfigHash` lockstep (sp1-contract-call#12), operator node-provisioning enforcement, and audit-as-a-unit with the slasher (#66). Merge stays a no-op for production — `Chain` is default and nothing selects `Unbounded` outside tests.

Excellent iteration — the determinism framing, the clamped-node diligence (finding the *partial* `Ok([Store])` case), and using the measured path as an oracle are all exactly right. Approving, subject to the standing cross-repo/audit gates and the pre-enablement handler re-measurement above.

BreadrichEngels:
Done — flipped my review to **approve** at the current head `b6d2d50`, after verifying the fix in code rather than trusting the summary.

**What I checked and confirmed:**
- **My blocking item (the calldata-gas omission) is genuinely fixed.** `estimate_applied_payload_gas` now prices the payload off the real encoded bytes: a **transport** term (EIP-2028 intrinsic calldata, 4/16) — the term I said was missing — plus a **dispatch** term (`2000 + 14·len` for the handler decode loop), returning `max(exec+transport, EIP-7623 floor)`. The two terms are distinct real costs (intrinsic vs in-handler CALLDATACOPY), so it's conservative, not double-counting, and the floor modeling matches protocol.
- **The oracle test is the right, durable closure** — and stronger than the single boundary test I'd asked for. `analytic_bound_dominates_measured_apply_cost` asserts `analytic ≥ measured(revm)` across sparse/dense stores, fat logs, and mixed shapes up to 650 updates, with the signature floor zeroed so no missing per-update term can hide. Combined with the gate, an accepted payload provably fits under `2²⁴`.
- **CI green** on all five checks; determinism guard (compile-time cold-SSTORE > warm) intact.

**On bagelface's two explicit asks:**
- *Naming* — I agreed to keep `Unbounded`; the `chainConfigHash` value-binding is a stronger guard than a type suffix. Told them not to re-open it.
- *Re-read the priced gate* — done; the soundness analysis carries over cleanly to the new invariant.

**One residual I folded into the pre-enablement gates (didn't block the merge):** the calibrated `14 gas/byte` constant is validated against *this analyzer's* handler, not the production `verifyAndUpdate` — so before `Unbounded` is enabled anywhere, those constants need re-measuring against the real SDK handler (ideally porting the oracle test into solidity-sdk). Plus three genuinely minor non-blocking notes and the unchanged standing gates (guest lockstep, node provisioning, audit-as-a-unit with #66).

Review posted: https://github.com/gas-killer/gas-analyzer/pull/166 (approved)

BreadrichEngels:
The 3rd comment is my own approval write-up — there's no new question from Ron. Nothing actionable here.

Status: PR #166 **merged by bagelface at the exact commit I approved** (`b6d2d50`) — my calldata-gas concern was fixed and verified, and my approval carried it in. Ron's mention is just a post-merge echo of the PR body.

I'm holding — no further GitHub comment. Posting again would only repeat my already-recorded review. The one open item (re-measuring the `14 gas/byte`/`2000` apply constants against the production `verifyAndUpdate` handler before `Unbounded` is enabled anywhere) is on record in my review as a pre-enablement gate, and merge is a no-op for production since `Chain` stays the default. If someone later moves to actually enable `Unbounded`, that's the point to pick the residual back up.

Issue #2, opened 2025-06-09
[Feat] Calculate gas savings


rubydusa:
Write a function `get_gk_savings` which computes how much gas would using gas killer save for a given function

Gas savings calculation is tricky because of many different factors affecting gas (intrinsic gas costs, calldata gas costs, memory and stack manipulation etc)

Easiest naive implementation is to have a deployed gk contract on testnet, simulate a transaction on it and return the diff

Issue #1, opened 2025-06-09
Research alternatives to forking testnet Anvil


rubydusa:
Forking a real testnet rpc endpoint in order to simulate a transaction is resource expensive.

- [ ] Research alternative means to produce a trace than GethAPI

Issue #162, opened 2026-06-24
Add heuristic-only dashboard mode to reduce RPC load


BreadrichEngels:
Add a heuristic-only mode for the dashboard to cut RPC consumption.

## Context

Dashboard RPC usage has been exhausting quotas — Alchemy maxed out, and the Sepolia dashboard hit its monthly quota and is currently disabled. Proposal: lean on heuristic estimates instead of full analysis for the dashboard.

## Viability

Partial — still need the trace for state changes, but the EVM sketch fork step is already saved (so part of the heavy RPC work is avoided regardless).

## Expected tradeoff

False positives are expected mostly around low-savings edge cases, not major bugs.

_From Gas Killer Weekly meeting — 2026-06-24._

PR #167, opened 2026-07-09
feat: historical gas-savings indexer service + BD dashboard

rubydusa:
## Summary

Adds the indexer workspace — a standalone service that continuously indexes an EVM chain, runs the gas-killer analysis on qualifying historical txs, and serves a BD dashboard. ~13.2k insertions; no changes to the existing gas-analyzer crates' logic.

- **New crates:** `indexer-rpc` (token-bucket rate limiter + retry), `indexer-store` (Postgres schema/migrations + rollup MVs), `indexer-resolver` (DefiLlama/Etherscan labeling), `indexer-api` (thin `Analyzer` wrapper over evmsketch/rpc), `indexer-service` (head-tracker / worker / refresher bins over Redis queue), `indexer-web` (axum + askama dashboard with auth, AI diagnostics).
- **Deploy:** production docker-compose stacks for mainnet + Sepolia (isolated compose projects), incl. an nginx **egress rpc-proxy** that rate-limits all upstream RPC (captures EvmSketch's internal provider traffic that bypasses the in-process limiter).
- **Docs:** `docs/INDEXER.md`, architecture diagram, AI diagnostics notes.

Currently running in prod (mainnet + Sepolia dashboards).

## Merge status vs main

~~Conflicting~~ **Resolved** in the merge commit `9d6e36f` (main merged into this branch):

1. **`Cargo.lock`** — took main's dependency resolution wholesale, re-added the indexer workspace members via cargo.
2. **#164 TURETZKY floor split** — `indexer-api` migrated to `TURETZKY_UPPER_GAS_LIMIT_BLS`. Kept the conservative BLS floor since the report schema has a single `gaskiller_gas_estimate` column; reporting the lower Schnorr floor (27k) alongside it is a schema change, left as a follow-up (relevant to #162).
3. **#147 trace-fetch optimization** — passes the already-fetched `receipt.status()` into `compute_state_updates_from_tx`; saves one duplicate receipt RPC per analysis.
4. **#124 msg.value forwarding** (found during the merge) — `estimate_state_changes_gas_with_preceding` gained a `tx_value` param; the analyzer now extracts `value` from the tx it already fetches for the function selector and forwards it.

Validated locally on the merged tree: `cargo check --workspace`, `cargo clippy --workspace --all-targets` (no warnings), `cargo fmt --all --check`, indexer crate unit tests all pass.

Draft pending CI + review of the follow-up decision in item 2.


rubydusa:
@bagelface I mostly want your opinion on the file structure and high level overview

bagelface (commented):
Opened PR to fix some minor organizational issues https://github.com/gas-killer/gas-analyzer/pull/174/changes

Other than that, I think there's just a sweep that needs to be done to clean up all the AI generated markdown files and such

BreadrichEngels (commented):
Reviewed by cloning the branch and reading the high-risk surfaces directly. This is a big, well-structured PR — the crate boundaries are clean, the nginx egress proxy is a genuinely clever way to catch EvmSketch's bypassed reads, and the merge-conflict resolutions (#124 `tx_value` forwarding, #147 `receipt.status()` reuse, #164 `TURETZKY_UPPER_GAS_LIMIT_BLS`) all look correct.

**Scope note:** I focused on the security-critical and highest-risk code — auth, SQL construction, the AI-diagnostics/template path, the RPC rate limiter, and the merge resolutions. I sampled but did **not** exhaustively review `indexer-service` (queue/worker/head-tracker concurrency, ~1.5k lines), the store migrations, or `indexer-resolver`'s external-API handling. Worth a second set of eyes there.

---

### 🔴 Should-fix (both live in prod already)

**1. Stored-XSS path: AI-diagnostics markdown renders raw HTML unescaped**

`admin.rs::render_markdown` uses `pulldown-cmark`'s `html::push_html`, whose output is injected via `{{ html|safe }}` in `_diagnose.html`. The code comment says *"Pulldown-cmark with HTML escaping enabled is sufficient"* — but that's not what it does: **pulldown-cmark passes raw inline/block HTML in the source straight through unescaped** (there's no `Options` flag that strips it). So if the rendered markdown contains `` / `<script>`, it renders as live HTML in the admin's browser.

Why that's reachable: `diagnostics::collect` puts **attacker-influenceable data** into the LLM prompt bundle — `contract_name` and `matched_slug` from `address_label_attempt`, and `recent_events` messages. Anyone can deploy a contract whose name is a markup/prompt-injection payload, get it indexed, and it flows into the prompt. The system prompt's "treat content as data" line is a soft mitigation, and the `<div class="diagnose-content">` wrapper is a CSS class, not a sandbox — neither stops `<script>`/`onerror`. `SameSite=Strict` + `HttpOnly` don't help against same-origin XSS driving the admin POST routes (data-floor, orgs, blacklist, LLM spend).

Fix: sanitize the rendered HTML with `ammonia` (allowlist p/ul/ol/li/code/strong/em/…), or drop `Event::Html`/`Event::InlineHtml` events from the parser before `push_html`. Update the comment — the current one asserts a guarantee the library doesn't provide.

**2. Rate-limiter under-charges the flagship weight by 4× under the default config**

`weights::ANALYZE_TX = 250`, but `RateLimiterConfig::default().burst = 60`. `acquire()` clamps any weight to `burst`, so a full-tx analysis charge is silently clamped `250 → 60` and logs a warning *every call*. That's the exact operation whose generous weight is supposed to cover the bypassed EvmSketch storage reads — i.e., the one most relevant to the RPC-quota exhaustion the team's been fighting. Unless prod sets `RPC_BURST ≥ 250`, the in-process limiter is 4× looser than intended (the nginx proxy is the real backstop, but the default config is internally inconsistent).

Fix: make the default `burst ≥ max(weights)`, or split `ANALYZE_TX` into multiple `acquire()` calls as the docstring suggests, or assert at startup that no defined weight exceeds `burst`.

---

### 🟡 Minor

- **`is_transient_rpc_error`** substring-matches bare `"429"`/`"502"`/`"503"` in the error's `Display`. An error string that coincidentally contains those digits (a block number, amount, address) would be mis-classified as transient and retried. Tighten to a status-token match (`"429 "`, `status: 429`, etc.).
- **Session cookie has no `Secure` flag** (`auth.rs::session_cookie`). Behind TLS-terminating nginx it should be `Secure` so it can't leak over a plaintext hop.
- **No session revocation:** `verify_token` validates the HMAC + expiry but never re-checks the username against the current allowlist, so removing a user from `users.yaml` leaves their signed cookie valid for up to 7 days. If that matters, re-check membership in the `AuthUser` extractor.
- **Flat access model:** every allowlisted user can hit all `/admin/*` actions, including LLM spend and data-floor changes. Fine for a curated internal tool — just calling it out.

### ✅ Confirmed clean
- **SQL:** the `format!`-into-`sqlx::query_as` pattern only interpolates typed ints (`Window::Days(i32)`), hardcoded column names, and a typed `NaiveDate` data-floor — all user values (`slug`/`address`/`selector`/filters/`chain_id`) are `.bind()`-ed. `TxFilters::where_clause` builds numbered placeholders and binds in order. No injection. (Nicely, the data-floor helper even documents why it's safe.)
- **Auth coverage:** every dashboard + admin handler takes the `AuthUser` extractor; only login/logout/healthz/static are open. HMAC compare is constant-time (`subtle`), bcrypt for passwords, `HttpOnly` + `SameSite=Strict`.
- **nginx egress proxy:** DELAY-mode `limit_req` + revm's one-at-a-time blocking reads = clean backpressure; streams trace bodies; SNI handled; upstream RPC key stays out of client-facing logs.

Nice work overall — the two should-fix items are both containable, and #1 is the one I'd prioritize given it's already serving prod.

BreadrichEngels (commented):
## Re-review (head `78ca172`) — all flagged items resolved

Re-reviewed the three fix commits since `d41fb9f5`. Every item I raised is fixed, and each landed with a regression test. `test-indexer` (green) now exercises all of them.

### 🔴 → ✅ Should-fix #1 — stored-XSS in AI-diagnostics markdown (`dafa6c3`)
`render_markdown` now downgrades raw HTML at the event level:
```rust
Event::Html(raw) | Event::InlineHtml(raw) => Event::Text(raw),
```
This is a complete fix — mapping to `Event::Text` forces `push_html` to escape the content, so both block (`<script>`) and inline (``) raw HTML render inert, while renderer-generated markup (`*world*` → `<em>`) is unaffected. Catching **both** `Html` and `InlineHtml` matters (pulldown-cmark split them), and you got both. The `markdown_raw_html_is_escaped` test pins `!contains("<script")` + `contains("&lt;script")` + working `<em>`. The docstring is now accurate too — it no longer claims a guarantee the library doesn't provide. I like that you dropped raw HTML entirely rather than reaching for `ammonia`; fewer moving parts and no allowlist to drift.

### 🔴 → ✅ Should-fix #2 — rate-limiter under-charging `ANALYZE_TX` (`b8bd0e8`)
`acquire()` now charges the full weight in `burst`-sized chunks:
```rust
let chunk = remaining.min(self.burst);
self.bucket.until_n_ready(chunk).await; remaining -= chunk;
```
This is the right fix and, crucially, **decoupled from config** — the flagship 250-token charge is now honored in full even under the default `burst=60` (60·4 + 10), pacing at `rps_budget` instead of silently clamping to 60. Since `chunk ≤ burst` always, `until_n_ready` can never hit its `N > burst` error path, so the `tracing::error!` branch is correctly just defensive. `heavy_weight_charges_fully_in_chunks` verifies the refill pacing. The in-process limiter is now internally consistent regardless of whether prod sets `RPC_BURST` (the nginx egress proxy remains the real backstop).

### 🟡 → ✅ Minor items
- **`is_transient_rpc_error` (`b8bd0e8`)** — now routed through `contains_status_code`, which requires non-hex-digit boundaries and rejects a `0x` prefix, so `"block 15029"`, `"exceeds 4290000"`, a hash containing `…a429f…`, and `"pc 0x429"` no longer misclassify as retryable, while `"429 Too Many Requests"` / `"status 502"` / `"(code: 503…)"` still do. The `transient_matches_status_tokens_not_embedded_digits` test covers exactly these. Nice — this is more careful than the simple `"429 "` token match I suggested.
- **Secure cookie + session revocation (`78ca172`)** — `session_cookie` and `clear_cookie` both set `Secure` now (with the correct note that browsers exempt `localhost` so dev logins still work), and `verify_token` re-checks allowlist membership after signature+expiry, so editing `users.yaml` + restart revokes outstanding cookies within the process lifetime rather than honoring them for the rest of the 7-day TTL. `token_rejected_after_user_removed_from_allowlist` locks it in.

### One residual, non-blocking (same threat class as #1)
The HTML-event downgrade neutralizes raw HTML, but markdown **link/image URI schemes** are still renderer-generated and pass through unfiltered — i.e. if the model were induced (via the attacker-influenced `contract_name`/`recent_events` in the prompt bundle) to emit `[x](javascript:alert(1))` or `data:text/html,…`, `push_html` would render a live `href`. Lower likelihood than raw `<script>` since it depends on steering the LLM's output, but it's the same attacker-data-to-admin-browser path. Cheap defense-in-depth: after parsing, drop or rewrite `Event::Start(Tag::Link/Image)` whose URL scheme isn't in an allowlist (`http`/`https`/`mailto`/relative). Worth a follow-up, not a blocker.

### Standing scope caveat (unchanged)
As on my first pass, this confirmation covers the security-critical surfaces — auth, SQL, the diagnostics/template path, the rate limiter, the merge resolutions. I still have **not** exhaustively reviewed `indexer-service` concurrency (queue/worker/head-tracker), the store migrations, or `indexer-resolver`'s external-API handling — those deserve a second reviewer, and rubydusa's own ask to @bagelface on file structure + the AI-markdown cleanup sweep (#174 and the doc/`*.md` pass) is still outstanding. From the surface I did review, the blockers are cleared and this is in good shape to merge once those remaining pieces are signed off. Thanks for the fast, precise turnarounds — every fix hit exactly what was flagged.

PR #168, opened 2026-07-09
feat: pinned code overlays for unbounded simulation (UNBOUNDED_V2)

RonTuretzky:
Stacked on #166 (merge that first; this diff is only the overlay commits).

## What

Extends the pinned unbounded env from `{gas limits}` to `{gas limits, address → code}`: large immutable blobs (LLM weights, lookup tables) mount as contract code in the simulation environment, verified against a single 32-byte manifest hash the consumer commits on-chain — no deployment, no on-chain directory.

Concretely, for the motivating consumer (solidity-sdk `onchain-llm`, Qwen3-0.6B): 597MB of weights = ~24.3k data contracts and ~130B gas to deploy in directory mode, vs **one immutable hash** in overlay mode. Measured end-to-end on anvil with the overlay mounted via setCode: 16-id chat prompt + 8-token answer ("Ethereum is a decentralized blockchain platform", bit-exact vs the integer reference) = **545.1B simulated gas** → one `Store` + logs on-chain.

## How

- **core/overlay.rs**: `CodeOverlay`, `OverlayEnv::from_blobs/verify` (chunk 24,575B → `0x00||payload`, derive addresses, hash; refuse manifest mismatches), `overlay_manifest_hash`, `overlay_chunk_address` (mirrors `Qwen3Engine.overlayChunkAddress`, cross-pinned by test vectors verified against the deployed Solidity via `cast call`), and `env_commitment` — the versioned pinned-env value (V1/V2 domain-separated) that **the SP1 guest must bind into `chainConfigHash`** (companion guest change, same pattern as #166's env overrides; overlay mode must not serve slashable quorums until that lands).
- **rpc**: `*_with_env` variants of the three `debug_traceCall` helpers mount the bindings as `stateOverrides {address: {code}}` alongside the V1 gas overrides; `*_with_profile` delegate with `None` (no API breaks).
- **evmsketch**: `call_to_encoded_state_updates_with_evmsketch_env` threads `Option<&OverlayEnv>` through the hybrid prestate/struct-log extraction; profiled/plain entrypoints unchanged.
- **Shape gate unchanged**: overlaid chunks are STOP-prefixed pure data — they can't execute or be written to, and payload `Store`s only target consumer storage.
- Payload gas estimation stays on the real chain env, as in #166.
- `docs/UNBOUNDED_OVERLAYS.md`: mechanism, derivation formulas (bit-identical to the consumer side), operator lifecycle, availability-is-liveness-not-safety.

## Tests

- core: derivation vectors cross-pinned against solidity-sdk, chunking/prefixing, manifest-mismatch refusal, env-commitment version/content separation (36 core tests pass).
- rpc/evmsketch: compile-threaded; existing unit tests pass (22 evmsketch lib tests). The anvil-crate integration tests fail identically on the base branch (pre-existing env issue, untouched here).
- Companion consumer + E2E: gas-killer/solidity-sdk branch `RonTuretzky/onchain-solidity-llm` — `GasKillerChat` overlay mode, `test_OverlayModeMatchesDirectoryMode` (vm.etch-mounted overlay produces byte-identical generation to directory mode), `tools/deploy_anvil.py --overlay`.

## Review focus (honest guesses)

- `StateOverridesBuilder`/`AccountOverride::with_code` usage matches alloy 1.0.37 — please sanity-check against your preferred override-construction style.
- `env_commitment` layout (domain || gas limits || manifest || n || (addr, codeHash)*) is a proposal — happy to align it with whatever the SP1 guest branch wants to hash.
- Overlay bytes currently arrive as in-memory blobs (`from_blobs`); a file-backed artifact store with paths in operator config is deliberately left to the service-side wiring PR.

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -7,6 +7,7 @@
 pub mod constants;
 pub mod encoding;
 pub mod heuristic;
+pub mod overlay;
 pub mod prestate;
 pub mod sim_profile;
 pub mod trace;
@@ -20,6 +21,11 @@ pub use heuristic::{
     BASE_TX_COST, LOG_BASE_COST, LOG_DATA_COST_PER_BYTE, LOG_TOPIC_COST, TraceOperations,
     WARM_SSTORE_COST, estimate_gas_from_operations, estimate_gas_from_state_updates,
     extract_operation_counts_from_trace,
+};
+pub use overlay::{
+    CodeOverlay, ENV_COMMITMENT_DOMAIN_V1, ENV_COMMITMENT_DOMAIN_V2, OVERLAY_CHUNK_PAYLOAD,
+    OVERLAY_DOMAIN_V1, OverlayEnv, OverlayError, env_commitment, overlay_chunk_address,
+    overlay_manifest_hash,
 };
 pub use prestate::{
     PrestateEligibility, build_state_updates_from_prestate, classify_prestate_eligibility,
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -31,7 +31,8 @@ pub use prestate::{
     PrestateEligibility, build_state_updates_from_prestate, classify_prestate_eligibility,
 };
 pub use sim_profile::{
-    SimProfile, UNBOUNDED_V1_BLOCK_GAS_LIMIT, UNBOUNDED_V1_TX_GAS_LIMIT, UnboundedShape,
+    SimProfile, UNBOUNDED_V1_BLOCK_GAS_LIMIT, UNBOUNDED_V1_TX_GAS_LIMIT,
+    UNBOUNDED_V1_XL_BLOCK_GAS_LIMIT, UNBOUNDED_V1_XL_TX_GAS_LIMIT, UnboundedShape,
     UnboundedShapeViolation, validate_unbounded_shape,
 };
 pub use trace::{compute_state_updates, copy_memory, parse_trace_memory};
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -8,6 +8,7 @@ pub mod constants;
 pub mod encoding;
 pub mod heuristic;
 pub mod prestate;
+pub mod sim_profile;
 pub mod trace;
 pub mod types;
 
@@ -22,6 +23,10 @@ pub use heuristic::{
 };
 pub use prestate::{
     PrestateEligibility, build_state_updates_from_prestate, classify_prestate_eligibility,
+};
+pub use sim_profile::{
+    SimProfile, UNBOUNDED_V1_BLOCK_GAS_LIMIT, UNBOUNDED_V1_TX_GAS_LIMIT, UnboundedShape,
+    UnboundedShapeViolation, validate_unbounded_shape,
 };
 pub use trace::{compute_state_updates, copy_memory, parse_trace_memory};
 pub use types::{
--- a/crates/core/src/overlay.rs
+++ b/crates/core/src/overlay.rs
@@ -196,7 +196,10 @@ pub fn env_commitment(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::sim_profile::{UNBOUNDED_V1_BLOCK_GAS_LIMIT, UNBOUNDED_V1_TX_GAS_LIMIT};
+    use crate::sim_profile::{
+        UNBOUNDED_V1_BLOCK_GAS_LIMIT, UNBOUNDED_V1_TX_GAS_LIMIT, UNBOUNDED_V1_XL_BLOCK_GAS_LIMIT,
+        UNBOUNDED_V1_XL_TX_GAS_LIMIT,
+    };
     use alloy_primitives::{address, b256};
 
     /// Cross-ecosystem pin: these vectors are reproduced by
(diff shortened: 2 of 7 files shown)


bagelface:
## Proposal: derive overlay chunk addresses with CREATE3, so the overlay and directory address spaces become one

Came at this from the other end: an external integrator deployed `GasKillerChat` in overlay mode against the published `qwen3-0.6b-onchain-v1` manifest (`0x23216cb9…7ae9`) and pointed the public testnet at it. It fails, of course, since this PR is unmerged. Details in gas-killer/service, but the interesting part for #168 is a design observation, not the failure.

### The observation

`overlay_chunk_address` (`crates/core/src/overlay.rs:118`) derives

```
address = keccak256("gaskiller.llm.overlay.v1" || manifestHash || uint64(index))[12:]
```

Those addresses have **no deployment preimage**. Landing code at one means a 160-bit search for a CREATE or CREATE2 preimage. So overlay mode and directory mode are permanently disjoint address spaces, which is why `Qwen3Engine._resolve` branches on `rootDirectory == address(0)` and why the solidity-sdk spec frames the two as an either/or ("choose per model: small models on-chain, large models overlaid").

That dichotomy is avoidable, and this PR is the cheapest moment to avoid it.

### The change

Derive the chunk address with CREATE3 against a pinned factory, keeping the same salt material:

```
salt   = keccak256("gaskiller.llm.overlay.v2" || manifestHash || uint64(index))
proxy  = keccak256(0xff || FACTORY || salt || keccak256(PROXY_INITCODE))[12:]
chunk  = keccak256(0xd694 || proxy || 0x01)[12:]
```

CREATE3 specifically, not CREATE2. A CREATE2 address depends on the init-code hash, which depends on the chunk payload, and the whole point is that the engine derives the address *before* it has the payload. CREATE3's address depends only on `(factory, salt)`, which is exactly the property required. This is the same reason 0xsequence/sstore2's `SSTORE2Map` uses CREATE3 rather than CREATE2, and `OVERLAY_CHUNK_PAYLOAD`'s docstring already notes the SSTORE2 lineage of the `0x00 || payload` runtime.

### What it buys

1. **One address space.** A chunk address is derivable from the manifest alone, so no on-chain directory is needed (the entire win of overlay mode is preserved), and the same address is now deployable by anyone.

2. **Progressive materialization.** A consumer commits to a manifest once. Operators mount overlays for chunks with no code at the reference block; anyone who wants durability deploys real chunks at those exact addresses, incrementally, permissionlessly, funded by whoever cares. No consumer redeploy, no manifest change, and the engine's read path is untouched, because the PR's own framing holds: an overlaid chunk is indistinguishable from a deployed data contract under `EXTCODESIZE`/`EXTCODECOPY`.

3. **A real answer to the availability objection.** `docs/UNBOUNDED_OVERLAYS.md:52` currently establishes that "availability is a liveness concern, never a safety one" and stops there. Under a shared address space the exposure becomes recoverable by anyone, without touching the consumer. For external operators that is a materially better story, and it may soften the gas-killer/service#356 conversation.

4. **Shrinking overlay transport.** Every chunk that gets deployed on chain drops out of the `stateOverrides` body. It does not rescue the RPC path (#169's ~1.2GB-per-call figure stands), but the two changes compose rather than conflict.

### What has to change

- `overlay_chunk_address`: CREATE3 derivation, new domain (`…overlay.v2`), plus `FACTORY` and `PROXY_INITCODE` hash as pinned constants. The derivation test vectors cross-pinned against the deployed Solidity have to be regenerated.
- solidity-sdk: `Qwen3Engine.overlayChunkAddress` mirrors it, so this spans both repos and the `RonTuretzky/onchain-solidity-llm` branch. The factory has to be deployed at a pinned address per chain, which is a new deployment dependency the current scheme does not have. That is the main cost of the idea and worth weighing.
- `env_commitment` (`overlay.rs:173`): add the reference block. The layout already carries `(address || keccak(code))` per chunk, so per-chunk authentication is present and an operator can check on-chain code at an address against the committed hash. What is missing is that *which source served a chunk* becomes block-dependent once both sources exist.
- Pin the mount rule explicitly: on-chain code at the reference block wins, the overlay fills gaps only. Every operator and the SP1 guest must enforce it identically. Divergence risk is low, since both sources yield bytes that hash to the same committed value, but the rule cannot be left implicit.
- The shape-gate argument is currently "overlaid chunks are STOP-prefixed pure data, so they cannot execute or be written to." Worth re-checking that it still reads the same way when a chunk address is a real deployable account rather than an unreachable phantom.

### Cost

CREATE3 adds a throwaway proxy per chunk, roughly 40k gas and one extra account. On the 0.6B model's ~129.3B gas full deployment that is under 1%, and it is only paid by whoever chooses to materialize a chunk. Nobody pays it in pure overlay mode.

For calibration on why "just deploy it" is not the alternative: 598.7MB at 200 gas/byte code deposit plus 16 gas/byte calldata is ~129.3B gas, which is 8,100 transactions under EIP-7825's 2^24 per-tx cap, and ~133 ETH at Sepolia's current 1.03 gwei base fee. The deposit is 92% of that and no library choice touches it.

### What this does not fix

Nothing about the integrator's failure. It does not affect the RPC gas cap, does not substitute for #169, and does not shorten the ~10 minute operator leg on a 545B-gas `ask()`. It is a design change to land inside this PR, not a shortcut around it.

Happy to open it as a diff against `ron/unbounded-v2-code-overlays` if the direction looks right. The factory-deployment dependency is the part I would want a second opinion on before writing code.

Issue #113, opened 2026-04-30
Detect fallback re-entry into target contract during `runStateUpdatesCall`


rubydusa:
## Problem

`StateChangeHandlerGasEstimator` (added in #101) is a transparent proxy: `runStateUpdatesCall` runs locally, every other selector is forwarded to the original implementation via `DELEGATECALL`. This is great for letting external protocols (oracles, AMMs, etc.) call back into `contract_address` and still get a valid response *while a simulation is in flight*.

The flip side: nothing currently tells us when that fallback path actually fired during a `runStateUpdatesCall` simulation. Concretely:

1. We invoke `runStateUpdatesCall(types, args)` against the proxy.
2. Replaying one of those state updates issues a `CALL` to some external contract.
3. That external contract calls back into our proxy address (oracle pull, ERC-777 hook, AMM router callback, etc.).
4. Because the inbound selector isn't `runStateUpdatesCall`, the proxy hands the call to the original implementation via `DELEGATECALL`.
5. The simulation finishes and we report a gas estimate — with no signal that re-entry happened.

This matters because:
- The gas number we produce conflates the cost of the state-update replay with whatever the implementation did inside the fallback. The implementation, in turn, may itself read storage that the GasKiller path is supposed to be replacing — so the reported number can drift from what a real on-chain GasKiller call would cost.
- Downstream consumers (the AVS, the CLI's coloured tag system) currently can't distinguish "estimate was clean" from "estimate included a fallback round-trip we don't understand". Both look like a measured number.

We need a way for the estimator to surface "a callback hit the proxy during this simulation" so we can either flag the report, treat the estimate as heuristic, or special-case the value.

## Prior art and why a proper fix is non-trivial

The obvious idea — set a storage flag in the fallback, read it at the end of `runStateUpdatesCall` — does not work in general, because the simulated call may be a `STATICCALL` (any external view path that re-enters us). `SSTORE` reverts under `STATICCALL`, so we'd either fail simulations that re-enter via a view function, or we'd have to detect the static context and skip flagging — i.e. silently miss exactly the case we care about.

[...]

PR #171, opened 2026-07-12
fix: heuristic accuracy — refunds, SSTORE pricing, calldata, re-entrancy flag

rubydusa:
Fixes the known heuristic discrepancies documented in `crates/core/src/heuristic.rs`, making heuristic-only mode (#170, #162) produce usable numbers. Originally stacked on `Rubydusa/162-heuristic-only-mode`; #170 has merged, so this now lands on `main`.

## The problem

The heuristic compares pre-refund, mispriced components against post-refund receipt `gasUsed`. On the validated MEV arb tx (`0x19daad7a…a16795`) it overshot by ~19%, clamping reported savings to 0% — which is what heuristic-only mode would have served for essentially every call-heavy DeFi tx.

## The fixes (one commit each)

- **EIP-3529 refunds**: the trace's final refund counter (verified present on live mainnet traces; 5.6k–59.7k in one sampled block) is netted out, capped at gross/5.
- **SSTORE pricing**: actual per-write charge from struct-log `gasCost` (cold 22,100 / warm 100 / dirty 5,000) instead of flat 5,000. `WARM_SSTORE_COST` is gone.
- **Calldata**: EIP-2028 cost over the ABI-encoded state updates the replay ships (~350k gas on the MEV tx; this bias pushed estimates *down*).
- **Fallback double-count**: `extract_operation_counts_from_trace` no longer counts nested SSTOREs/LOGs whose gas is already inside `external_call_gas`.
- **Re-entrant callbacks — flagged, not corrected** (policy decision): a callee calling back into the origin contract (Uniswap swap callbacks; 72/128 depth-1 calls on the MEV tx) still counts as external gas, but `TraceExtract::reentered` is set — detected via a stack peek on CALL-family opcodes at depth ≥ 2, STATICCALL included (the view-re-entry case from #113). Surfaced as a new `reentered` column on `analysis` (auto-migration), a CLI warning, and `reentered` + optional `origin_address` in the wasm API (backward-compatible).

Also: `compute_state_updates` now returns a `TraceExtract` struct instead of a 4-tuple.

## Reconciliation with main (merge commit)

Main moved substantially while this was in flight; the merge commit reconciles the `TraceExtract` refactor with canonical checkpoints (#173), prestate hybrid extraction (#163/#165), and unbounded sim mode (#166):

- `extract_state_updates_hybrid` keeps its `(updates, skipped, call_gas)` tuple; its legacy struct-log arm adapts to the new `compute_state_updates` signature with re-entry detection off — the simulation path never surfaces the flag, only the historical-tx analyzers do.
- `compute_state_updates_canonical` stays tuple-returning: the heuristic aggregates `TraceExtract` carries have no consumer on the canonical path (its fallback estimates are operations-based).
- `UNBOUNDED_COLD_SSTORE_COST`'s compile-time guard against the removed `WARM_SSTORE_COST` is folded into its doc: the heuristic now charges traced actual SSTORE costs, which EIP-2929/2200 cap at exactly that constant, so the gate's bound holds by construction.

## Verification

- Live end-to-end on a mainnet Universal Router swap (`0xc6b65088…7492`): `reentered=true` with origin supplied / `false` without, and heuristic estimate **120,084 vs actual gasUsed 126,564** (~5% under, vs the old systematic overshoot).
- clippy `-D warnings` across all CI configs (workspace, wasm, anvil, cli+anvil); 8 new unit tests (refund cap, SSTORE gas summing, re-entry positive/negative).

## Not included

- `indexer-web` stores `reentered` but doesn't display it yet.
- The wasm dashboard JS must pass `tx.to` as `origin_address` to activate detection there.
- EIP-7623 floor pricing (replays execute enough that standard token pricing applies).
- A depth-1 self-call (the origin contract CALLing its own address) has the same mispricing as a callback but is not flagged — detection starts at depth ≥ 2.


rubydusa:
Superseded by #183 (identical branch and content, rebased base). The stack machinery pinned this PR to the merged-and-reverted `Rubydusa/162-heuristic-only-mode` branch and rejected every retarget — see #182 for the parent re-land.

PR #173, opened 2026-07-12
feat(core): canonical state-checkpoint encoder for re-entrant Gas Killer tasks

RonTuretzky:
## Why

When a Gas Killer task's native execution makes an **external call** (and especially when that external contract **re-enters** the Gas Killer contract), the on-chain settlement must present the target contract's storage exactly as native execution would have at that moment — otherwise a re-entrant read/write observes state that never existed and the transition corrupts.

The legacy encoder (`compute_state_updates`) emits raw depth-1 `SSTORE`s in trace order and drops nested writes, leaning entirely on each emitted `CALL` **replaying natively** to reproduce them. That's fragile: it makes no guarantee about the storage an external re-entrant call sees mid-program, and if a re-entrant call's on-chain replay diverges from the simulation the final state silently breaks. (It also emits reverted-`DELEGATECALL` writes as ghost `Store`s.)

## What

`compute_state_updates_canonical(trace, target)` tracks the **target's storage image across all frames** with revert-aware journaling (exactly like the EVM: child journals commit into the parent on success, drop on revert) and emits it as **canonical state slices**:

1. **Before every emitted external `CALL`/`CREATE`/`CREATE2`** — a slice brings the target's storage up to the exact image native execution had at that point. So external code that runs during that call — including a re-entrant call back into the target — reads canonical state. *This is the feature: the contract switches into the canonical intermediate state before making the external re-entrant call.*
2. **A final slice** — re-asserts every touched slot's native-final value, so the committed end state is canonical **even if a re-entrant call's on-chain replay diverged**. When replay matched, these are cheap same-value writes.

Re-entrant writes (a callee writing back into the target) are journaled into the canonical image but **not** emitted at their nested position — they replay natively during the `CALL` — and get re-asserted by the next slice, so they land canonically regardless.

### No contract change

The output is still an ordinary `(StateUpdateType[], bytes[])` program of the **existing** `STORE`/`CALL`/`LOG`/`CREATE` update types. The on-chain `StateChangeHandlerLib` already applies these in order, so canonical checkpointing is purely a smarter **ordering + slicing** of primitives the contract already supports — it drops in with zero Solidity changes. (It also composes with, and largely supersedes the need for, the blunt re-entrancy *guard* — instead of blocking re-entrancy, it makes re-entrancy observe correct state.)

### Determinism

Within a slice, slots are emitted in ascending order and consecutive writes to the same slot between two boundaries collapse to the boundary value (intermediate values are unobservable on-chain — no external code runs between boundaries). So independent provers produce byte-identical programs → identical task digests.

### Secondary correctness fixes (tested)

- **Reverted `DELEGATECALL` writes are discarded** (journaling drops a failed frame's writes; legacy emits them).
- **Writes by an unrelated callee are never attributed** to the target's canonical image.

## Interface

- `gas_analyzer_core::compute_state_updates_canonical` — the encoder.
- evmsketch: `StateEncoding::{Legacy, Canonical}` + `call_to_encoded_state_updates_with_evmsketch_mode(...)`. The existing `call_to_encoded_state_updates_with_evmsketch(...)` is unchanged (Legacy), so nothing regresses until a caller opts in.

## Tests

11 new unit tests over hand-built traces: final-slice sort/collapse, pre-call slice visibility, re-entrant write re-assertion, re-entrant read between two calls seeing canonical state, reverted vs committed DELEGATECALL, errored CALL not emitted, unrelated-callee isolation. `cargo test -p gas-analyzer-core` green; core clippy + fmt clean.

The service side (selecting Canonical behind a config flag, node + router in lockstep so digests agree) lands in a companion service PR.

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -20,7 +20,9 @@ pub use heuristic::{
     WARM_SSTORE_COST, estimate_gas_from_operations, estimate_gas_from_state_updates,
     extract_operation_counts_from_trace,
 };
-pub use trace::{compute_state_updates, copy_memory, parse_trace_memory};
+pub use trace::{
+    compute_state_updates, compute_state_updates_canonical, copy_memory, parse_trace_memory,
+};
 pub use types::{
     DummyExternal, IStateUpdateTypes, Opcode, RevertingContext, SimpleStorage, StateUpdate,
     StateUpdateType, StateUpdates,
--- a/crates/core/src/trace.rs
+++ b/crates/core/src/trace.rs
@@ -1016,6 +1016,42 @@ mod tests {
     }
 
     #[test]
+    fn canonical_total_call_gas_matches_legacy_for_delegatecall_trace() {
+        // TARGET delegatecalls a library, which (still in TARGET's context)
+        // makes an emitted CALL to y. The delegatecall itself is never emitted
+        // and never charged in either encoder — only the CALL is — so both
+        // encoders must agree on total_call_gas despite Canonical hoisting
+        // delegatecall frames instead of emitting them.
+        let lib = Address::new([0x33; 20]);
+        let y = Address::new([0x44; 20]);
+        let logs = vec![
+            delegatecall(1, 80_000, lib, None),
+            call(2, 70_000, y),
+            resume(2, 65_000, true),
+            resume(1, 60_000, true),
+        ];
+        let canonical_trace = DefaultFrame {
+            failed: false,
+            gas: 0,
+            return_value: Default::default(),
+            struct_logs: logs.clone(),
+        };
+        let (_, _, canon_gas) =
+            compute_state_updates_canonical(canonical_trace, TARGET_ADDR).expect("canonical");
+        let legacy_trace = DefaultFrame {
+            failed: false,
+            gas: 0,
+            return_value: Default::default(),
+            struct_logs: logs,
+        };
+        let (_, _, legacy_gas) = compute_state_updates(legacy_trace).expect("legacy");
+        assert_eq!(
+            canon_gas, legacy_gas,
+            "total_call_gas must agree between encoders on a delegatecall-heavy trace"
+        );
+    }
+
+    #[test]
     fn write_by_unrelated_callee_is_not_attributed_to_target() {
         // Inside CALL(x), x writes ITS OWN storage (storage_ctx = x ≠ TARGET).
         // That must never enter TARGET's canonical image.
--- a/crates/core/src/trace.rs
+++ b/crates/core/src/trace.rs
@@ -4,9 +4,9 @@
 //! Geth-format transaction traces (`DefaultFrame`). Contains only
 //! pure computation functions - no async, no I/O, no RPC calls.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet};
 
-use alloy_primitives::Address;
+use alloy_primitives::{Address, B256};
 use alloy_rpc_types::trace::geth::{DefaultFrame, StructLog};
 use anyhow::{Result, bail};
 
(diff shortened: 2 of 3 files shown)


BreadrichEngels (approved):
## Review: canonical state-checkpoint encoder for re-entrant tasks

**Verdict: Approve.** This is a careful, well-tested, purely-additive encoder. I walked the frame/journal state machine against the EVM-journaling model it claims to mirror and the core invariants hold; it's opt-in (`StateEncoding::Legacy` default), the legacy path is byte-for-byte untouched, and there's no Solidity change. All four checks are green (`test-native`, `test-wasm`, `lint`, benchmark — the latter actually shows `compute_state_updates` at `-3.0%`, no regression). My one substantive ask is about *how the service flips this on*, not about the code here — see the gate at the bottom.

### What I verified in `compute_state_updates_canonical`

**SSTORE attribution is context-correct.** A slot is journaled only when `cur_ctx == Some(target) && !op_errored`. `storage_ctx` propagates exactly as the EVM does: root = target; `DELEGATECALL`/`CALLCODE` inherit the parent's ctx (execute in the caller's storage); `CALL`/`STATICCALL` = callee; `CREATE*` = `None` (a fresh deployment can't alias an already-deployed target). So a re-entrant `CALL` back into the target journals (ctx = target) while an unrelated callee writing its own slot does not — matching `write_by_unrelated_callee_is_not_attributed_to_target`.

**Journaling is revert-aware and mirrors the EVM.** On a successful resume the child's `journal` merges into the parent (deeper writes win) and its buffered `out` is appended; on failure the whole sub-frame (journal + out + view) is dropped. That's precisely why `reverted_delegatecall_writes_are_discarded` passes and the legacy ghost-`Store` bug goes away, while `committed_delegatecall_writes_are_emitted` still lands.

**The `flush_slice` invariant is sound.** `canonical_visible_image` merges every open frame's journal bottom-up, and `flush_slice` only runs when the current frame is emitting. I confirmed emitting frames form a contiguous prefix from the root (emitting only propagates through `DELEGATECALL`/`CALLCODE`, which also inherit ctx; any `CALL`/`STATICCALL`/`CREATE` child is non-emitting and *all* its descendants are non-emitting). Therefore at every flush all open frames are emitting **and** target-scoped, so the merged image is exactly the target's native storage at that boundary. Emitting a slice before each external `CALL`/`CREATE` (so re-entrant code reads canonical state) plus a final slice (so the committed end state is canonical even if a re-entrant replay diverged) is the right shape, and the `emitted_view` same-value suppression keeps slices minimal. `reentrant_read_between_two_calls_sees_canonical_image` nails the headline case.

**Determinism holds.** `BTreeMap` iteration + ascending-slot slices + collapse-to-boundary-value means independent provers emit byte-identical programs → identical digests. This is the property the whole settlement scheme leans on, and the encoder gets it structurally rather than by post-sort.

**Frame bookkeeping is defensive in the right places.** The pending-frame resolution (open iff `depth == parent_depth + 1`, else account inline) correctly distinguishes real frames from EOA/precompile/failed calls; the `bail!` on a depth jump `!= depth + 1` refuses to silently mis-attribute outcomes; gas is accounted exactly once per emitted call (at pop if a frame opened, inline otherwise) and relies only on the global monotonic gas counter, which is valid. `unreachable!`/`debug_assert!` on the CALL/CREATE/LOG paths are correct given `append_state_update_from_struct_log` only returns `Some` for `SELFDESTRUCT`/`TSTORE`.

### The gate before this can drive real settlement

This code is safe to merge, but it is **digest-affecting** the moment a caller passes `StateEncoding::Canonical`. Before the companion service PR flips any prover to Canonical:

1. **Strict node+router lockstep.** Every prover in a quorum must run the *same* encoder version, or two honest provers produce different programs → different digests → no quorum. The rollout must be all-or-nothing per network, not a rolling per-pod flip. (Same lockstep discipline as the commonware `Cargo.lock`-bump PRs.)
2. **Differential + soak on real traces before the cutover.** Every test here is a hand-built synthetic trace, so correctness currently rests on assumptions about geth `structLog` semantics (resume log carries the child success flag on stack top; depth/gas monotonicity). Please add at least one real mainnet re-entrancy trace run through both encoders with an on-chain replay equivalence check, and a soak comparing Legacy vs Canonical digests across a corpus, before enabling it. I'd treat green there — plus Ron's domain sign-off — as the merge gate for the *service* PR, not this one.

### Non-blocking nits
- `total_call_gas` accounting under Canonical is plausibly not identical to Legacy for delegatecall-heavy traces (delegatecalls are hoisted, not emitted as `CALL`s, and carry no `emitted_call_gas`). It doesn't affect program correctness (gas is only the estimate), but if any caller compares the estimate across encodings, worth a one-line test pinning expected `total_call_gas` on a delegatecall trace.
- Consider a fuzz/property test asserting the two structural invariants directly on random valid frame trees: (a) final committed image == full journal merge, and (b) every emitted slot value equals the canonical image at its boundary. That would guard the hot path against future refactors far better than the enumerated cases.

Excellent write-up in the PR description too — the "switch into canonical state *before* the external call" framing made the invariant easy to check. Nice work.

bagelface:
Re the two non-blocking nits in [this review](https://github.com/gas-killer/gas-analyzer/pull/173#pullrequestreview-4744721576):

**`total_call_gas` on delegatecall-heavy traces** — checked this empirically (Canonical vs Legacy on a delegatecall→CALL trace) and they agree (`5000 == 5000`). Neither encoder ever charges gas for the `DELEGATECALL` opcode itself, only for emitted `CALL`s, and both use equivalent "is this CALL nested in a filtered/non-emitting context" logic, so hoisting doesn't move `total_call_gas`. Added `canonical_total_call_gas_matches_legacy_for_delegatecall_trace` in 57ed758 to pin that.

**Fuzz/property test on random frame trees** — agree it'd be a better guard than the enumerated cases, but it's a bigger lift than a nit: there's no `proptest`/`quickcheck` dev-dependency in the workspace yet, so it'd mean pulling one in plus writing a valid-frame-tree generator and the two invariant checkers from scratch. Punting on this for now — happy to pick it up as a follow-up if useful.

PR #179, opened 2026-08-07
ci: lint every feature configuration with cargo-hack

bagelface:
## Problem

A single `cargo clippy` run only ever sees **one** feature configuration. Anything behind a
`#[cfg(feature = ...)]` that happens to be off in that run is never linted — so `-D warnings` is a
weaker gate than it looks.

The workspace has four such regions today, and three of them were unlinted:

| configuration | linted before this PR? |
|---|---|
| `gas-analyzer-evmsketch`, default | ✅ |
| `gas-analyzer-evmsketch` + `anvil` | ✅ (added in #165) |
| `gas-analyzer-evmsketch` *without* `anvil` | ❌ |
| `gas-analyzer-evmsketch` + `heap-profile` (6 gates in `benches/flamegraph.rs`) | ❌ |
| `gas-analyzer-cli` + `anvil`, no default features (2 gates in `main.rs`) | ❌ |

The last row is not hypothetical. That is the exact configuration CI already builds in
`cargo test (anvil)`, and it has been emitting **four warnings** into the job log for some time —
they just never failed anything, because that step has no `-D warnings`:

```
warning: unused variable: `tx_index`
warning: unused variable: `tx_sender`
warning: unused variable: `tx_value`
warning: function `decode_reverting_context` is never used
```

## Change

**`cargo hack clippy --each-feature`** over the two crates that declare features. It lints each
feature in both directions and picks up new features automatically, rather than depending on someone
remembering to enumerate them. Installed prebuilt via `taiki-e/install-action@cargo-hack`.

`--exclude-no-default-features` skips the all-features-off build of `gas-analyzer-cli`: with neither
`evmsketch` nor `anvil` the binary has no analysis backend at all, so it is not a configuration
anyone runs. Every other configuration passes.

**Fixed what it surfaced** in `crates/cli/src/main.rs`: `decode_reverting_context`, three `let`
bindings, and the three imports that fed them are all reachable only from the `evmsketch` path, so
they now carry `#[cfg(feature = "evmsketch")]` — matching the gate already on
`use alloy_eips::BlockNumberOrTag` a few lines below. No behaviour change in any configuration that
was already building.

## Verification

Each configuration `--each-feature` will run, checked locally with `-D warnings`:

| configuration | before | after |
|---|---|---|
| `cli` default | pass | pass |
| `cli --no-default-features --features anvil` | **4 warnings** | pass |
| `cli --no-default-features --features evmsketch` | pass | pass |
| `evmsketch` default | pass | pass |
| `evmsketch --features heap-profile` | pass | pass |
| `evmsketch --features anvil` | pass | pass |

## Cost

The `cli + anvil` configuration pulls `gas-analyzer-anvil` into the **lint** job, which currently
excludes it — about 3 minutes cold locally, amortised by `Swatinem/rust-cache` after the first run.
The `evmsketch` configurations are nearly free, since `anvil`/`heap-profile` are dependency-free
`[]` features and only that one crate recompiles.

## Note on #166

#166 replaces `evmsketch`'s `anvil` feature with `#[ignore]` on those tests, for the same reason this
PR exists: a feature that guards only test code buys nothing but a configuration clippy cannot see.
That removes one row from the matrix here. The two PRs touch different lines of the workflow and do
not conflict in either merge order.

## Side effect: `gas-analyzer-anvil` is now linted

`cargo clippy -p gas-analyzer-cli --all-features` builds `gas-analyzer-anvil` as a workspace path
dependency, and `RUSTC_WORKSPACE_WRAPPER` applies clippy to workspace members even when they are
built as dependencies. So that crate is now gated by `-D warnings` despite appearing in the
`--exclude` list of the older `cargo clippy (default workspace)` step.

That surfaced exactly one lint — `useless_borrows_in_formatting` in `crates/anvil/src/lib.rs:542` —
fixed here. Worth knowing rather than being surprised by later: `--exclude` on one step does not keep
a crate out of another step that pulls it in as a dependency.

## Toolchain note

CI runs clippy **1.97.0**; `useless_borrows_in_formatting` does not exist in 1.94, so that lint could
not be reproduced locally and was fixed from the CI output. Local `cargo clippy` is therefore not
authoritative for this workflow — new lints land with each stable release, and this job widens the
surface they apply to.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -31,6 +31,25 @@ jobs:
             --features gas-analyzer-evmsketch/anvil \
             --all-targets -- -D warnings
 
+      # A single clippy run only ever sees one feature configuration, so anything behind a
+      # `#[cfg(feature = ...)]` that is off in that run is never linted. `--each-feature` lints every
+      # feature of the crates that have them, in both directions, and picks up new features
+      # automatically instead of relying on someone remembering to enumerate them here.
+      #
+      # `--exclude-no-default-features` skips the all-features-off build of `gas-analyzer-cli`: with
+      # neither `evmsketch` nor `anvil` the binary has no analysis backend at all, so it is not a
+      # configuration anyone runs.
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@cargo-hack
+
+      - name: cargo clippy (feature matrix)
+        run: |
+          cargo hack clippy \
+            -p gas-analyzer-evmsketch \
+            -p gas-analyzer-cli \
+            --each-feature --exclude-no-default-features \
+            --all-targets -- -D warnings
+
       - name: cargo check (default workspace)
         run: |
           cargo check --workspace \
--- a/crates/anvil/src/lib.rs
+++ b/crates/anvil/src/lib.rs
@@ -539,7 +539,7 @@ pub async fn gas_estimate_block(
     println!("got {} receipts for block {}", receipts.len(), block_number);
     let mut reports = Vec::new();
     for receipt in receipts {
-        println!("processing {}", &receipt.transaction_hash);
+        println!("processing {}", receipt.transaction_hash);
         reports.push(
             get_report(&provider, receipt.transaction_hash, &receipt, &gk)
                 .await
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1,9 +1,12 @@
+#[cfg(feature = "evmsketch")]
 use alloy::consensus::Transaction as _;
+#[cfg(feature = "evmsketch")]
 use alloy::sol_types::SolError;
 use alloy::{hex, providers::ProviderBuilder};
 use alloy_provider::Provider;
 use anyhow::Result;
 use colored::Colorize;
+#[cfg(feature = "evmsketch")]
 use gas_analyzer_core::RevertingContext;
 use std::env;
 use url::Url;
@@ -13,6 +16,7 @@ use url::Url;
 /// Looks for hex-encoded revert data in the error message (the format revm
 /// produces: "Gas estimation reverted (gas: N): 0x...") and attempts to
 /// ABI-decode it as a `RevertingContext`.
+#[cfg(feature = "evmsketch")]
 fn decode_reverting_context(e: &anyhow::Error) -> Option<RevertingContext> {
     let msg = format!("{e:?}");
     let hex_start = msg.rfind("0x")?;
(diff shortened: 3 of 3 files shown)


BreadrichEngels (approved):
## Review: lint every feature configuration with cargo-hack

**Approve.** Correct diagnosis, right tool, and the fix is CI-verified by construction. Low risk — this is CI + dead-code gating with no runtime behavior change.

### The reasoning is sound
`cargo clippy` only ever sees one feature configuration, so `-D warnings` silently skips anything behind an *off* `#[cfg(feature = …)]` — and the `cli --no-default-features --features anvil` config was already emitting four warnings into the `cargo test (anvil)` log (no `-D warnings` there to catch them). `cargo hack clippy --each-feature` is the right instrument: it lints each feature in both directions and auto-discovers new ones, rather than a hand-maintained enumeration that drifts. Scoping it to the two feature-declaring crates and `--exclude-no-default-features` (a `cli` with no backend is a config no one runs) is the correct minimal surface.

### The gating fix is correct, and the CI step proves it
The `#[cfg(feature = "evmsketch")]` additions in `cli/src/main.rs` (`decode_reverting_context`, the three imports, `tx_index`/`tx_sender`/`tx`/`tx_value`) are safe precisely because those bindings were reported *unused* in the anvil-only config — i.e. the `#[cfg(feature = "anvil")]` block doesn't consume them, so removing them from that config can't break it. The strongest evidence is that the new `--each-feature -- -D warnings` step is green: every configuration of `cli` and `evmsketch` now both compiles and lints clean, so the gating is validated across the matrix rather than argued. The `anvil/src/lib.rs` `useless_borrows_in_formatting` fix (`&receipt.transaction_hash` → `receipt.transaction_hash`) is a no-op for `Display`.

The writeup's two incidental findings are both worth having surfaced: `gas-analyzer-anvil` getting linted via the `cli --all-features` dependency pull-in (an `--exclude` on one step doesn't hold across a step that pulls the crate in transitively), and the clippy-1.97-in-CI vs 1.94-local gap (local clippy isn't authoritative for this job). Good to have those documented rather than rediscovered.

### One very minor, forward-looking note
`--each-feature` covers each feature individually (+ default + no-default-features) but not feature *combinations* — it won't exercise a hypothetical `#[cfg(all(feature = "a", feature = "b"))]`. That's fully sufficient today (there are no `cfg(all(...))` gates, and the anvil-only run already catches the only cross-feature reference risk here), so no action needed — just worth knowing that if a `cfg(all(...))` region is ever added, it'd want `--feature-powerset` (exponential) or an explicit combo step to be covered. Not a gap in this PR.

Clean hardening — it closes a real `-D warnings` blind spot and the diff is self-verifying under the new job.

Issue #119, opened 2026-04-30
Investigate EvmSketch fork setup latency (block anchor initialization)


bagelface:
## Parent Epic
gas-killer/roadmap#9

## Context
From the epic: *"Validator forks blockchain at specific block height for each analysis — fork setup is a latency bottleneck."*

`GasKillerEvmSketch::build()` calls into `sp1-cc-host-executor` (BreadchainCoop fork, `cancun-v1` branch) to anchor state at a specific block. This cost is paid **per analysis request** because the block changes with each tx. The internal work likely includes fetching the block header, state proof data, and constructing the anchored context — all via RPC.

## Investigation Goals
1. **Measure** how long `build()` takes vs. `estimate_state_changes_gas()` — establish latency budget split
2. **Profile** what `sp1-cc-host-executor` does during construction (RPC calls, hashing, merkle proof verification)
3. **Evaluate** whether the `EvmSketch` context can be safely cached/reused across requests at the same block height (multiple txs from one block)
4. **Evaluate** whether `build()` can be pipelined — start it for the next block speculatively while the current analysis runs

## Deliverable
Comment in this issue with:
- Latency breakdown: fork setup vs. EVM execution vs. RPC reads (numbers from flamegraph or instrumented run)
- Whether same-block caching is safe given the `EvmSketch` API contract
- Specific follow-up optimization tasks with effort estimates

## Unknowns / Risks
- `sp1-cc-host-executor` is an external dependency — changes may require upstream PRs to BreadchainCoop
- The `cancun-v1` branch may be under active development (ABI instability risk)
- Caching the fork context may not be safe if it holds mutable state or open RPC connections
- **High timeline risk**: if the bottleneck is deep in sp1-cc, the fix may be out of scope for the 17-day sprint — investigation output will determine whether to pursue or defer

rubydusa:
## Investigation results — issue #119

### TL;DR

`build()` is **0.1–1.1%** of the per-request budget — basically noise.
The dominant cost depends on the analyzed tx's position in its block:

- **Early in block** (small `tx_index`): `debug_traceTransaction` dominates
  (~83%, ~8–11 s).
- **Late in block** (large `tx_index`): `replay_preceding_transactions`
  dominates (~91%, scales with `tx_index`; ~104 s for 69 preceding txs
  in the test sample).

Either way, the issue's premise — that `build()` is the bottleneck — is
incorrect. Build is two `eth_getBlockByNumber` calls plus an
[...]

PR #182, opened 2026-08-13
feat: heuristic-only analyzer mode to cut RPC usage ~300x

rubydusa:
> **Lineage**: re-lands #170, which was merged (`0af7474`) and then reverted on main (`50d117f`). GitHub cannot reopen a merged PR, so this carries the identical tree as a single revert-of-revert commit — including the original merge's reconciliation with #166. Merge this before #183, which builds on this content.

## What

Opt-in `HEURISTIC_ONLY` mode for the indexer worker: estimate gas from the extracted state updates alone, skipping the preceding-tx replay and the EvmSketch fork build entirely. Implements #162.

Also carries the two operational pieces that make the mode useful in practice: a matching reduced rate-limiter weight, and staleness controls (queue-job TTL, bounded head-tracker lag) for chains where analysis capacity trails block production.

## Why this cuts RPC ~300x

Per the measurements in #119, a full analysis of a mid-block tx costs ~1,593 RPC calls:

- ~1,556 from replaying the preceding transactions (`eth_getStorageAt` ×788, `getBalance`/`getCode`/`getTransactionCount` ×266 each, lazily fetched by the replay)
- ~30 more from the EvmSketch fork context setup
- ~5 for the analysis itself (receipt, block, tx, `debug_traceTransaction`)

Heuristic-only mode keeps just the last bucket. The `debug_trace` cannot be dropped — state-update extraction needs it — so this is ~5 calls per analysis, not zero.

## Tradeoff

The heuristic overshoots on call-dominated transactions (it compares pre-refund call gas against post-refund receipt gas), clamping their reported savings to 0%. Rows are flagged `is_heuristic = true` with `failure_reason = NULL`, so heuristic rows stay distinguishable in the dashboard. #171 addresses the heuristic accuracy itself.

## Rate-limiter weight

Heuristic-only analyses charge `HEURISTIC_ANALYZE_TX = 150` instead of `ANALYZE_TX = 250` — at the full weight the limiter would pace the cheap mode as if it were expensive, eating most of the quota win. The reduction is bounded by `debug_traceTransaction` itself (~122 weight units), which heuristic mode cannot drop: ~300× fewer calls ≠ 300× less compute-unit weight.

## Staleness controls

Heuristic-only exists to run high-volume chains on a small quota, which is exactly the regime where enqueue rate can permanently outpace analysis capacity. Two bounds keep analyses near head instead of arbitrarily stale:

- `QUEUE_JOB_TTL_SECS` (default `3600`): workers drop queued jobs older than this at claim time, checked before spending rate-limiter budget. Requeues keep the original enqueue timestamp so the TTL spans retries; legacy payloads without the field never expire.
- `MAX_BLOCKS_BEHIND` (default `0` = never drop): when the head-tracker cursor falls more than this many blocks behind head, it skips forward and drops the intermediate blocks instead of grinding through an ever-staler backlog. Sepolia compose defaults to 300 (~1h of blocks at 12s).

## Wiring

- `AnalyzerConfig.heuristic_only` (default `false`) → branch in `analyze_tx` before steps 5–6
- `--heuristic-only` / env `HEURISTIC_ONLY` on `indexer-service` (worker only)
- `HEURISTIC_ONLY` in `docker-compose.yml`, `SEPOLIA_HEURISTIC_ONLY` in `docker-compose.sepolia.yml` (namespaced like the rate knobs so both stacks toggle independently from the shared `.env`); same pattern for `QUEUE_JOB_TTL_SECS` / `MAX_BLOCKS_BEHIND`
- Docs (`INDEXER.md`, `ARCHITECTURE.md`) + `.env.example`

Env values must be literal `true`/`false` — clap rejects `0`/`1` with a clear error (verified empirically, including that compose's always-set `HEURISTIC_ONLY=false` default parses as false rather than "present = true").


BreadrichEngels (approved):
## Review — heuristic-only analyzer mode

**Approve.** Clean, well-scoped, and correctly on the indexer/reporting side — the `heuristic` estimator here is the dashboard path that the consensus gate (#166) deliberately does *not* share with, so nothing in this touches operator digest extraction or fraud-proof determinism. That keeps the stakes at "indexer correctness + RPC budget," which is exactly what this delivers. CI green including the benchmark. One process question on the lineage and a couple of minor notes; none blocking.

### The branch logic is right
The heuristic-only fork in `analyze_tx` correctly distinguishes the two heuristic origins that the dashboard depends on: intentional heuristic-only reports `(heuristic + BLS_floor, is_heuristic = true, failure_reason = None)`, while the pre-existing *measured-path-failed* fallback still reports `Some(truncated_reason)`. So "cheap mode" stays distinguishable from "the measured path errored" — the `failure_reason = NULL` invariant the PR claims holds in the code. The default path is moved verbatim into the `else` with no behavior change (confirmed the measured branch is byte-for-byte the same, just re-indented), and both branches add `TURETZKY_UPPER_GAS_LIMIT_BLS` consistently.

### The rate-limiter refactor is correct
Moving `limiter.acquire(analyze_weight)` out of `handle` and into the loop — *after* the TTL check — is the right ordering: an expired job is dropped before it spends any rate-limiter budget, which is the whole point of a staleness bound on a quota-starved worker. The permit's RAII scope still covers the `handle` call, so nothing changed about how long the budget is held. `HEURISTIC_ANALYZE_TX = 150` vs `250` is well-reasoned — bounded below by `debug_traceTransaction`'s own ~122 weight, which heuristic mode can't drop, so it correctly doesn't claim a 300× *weight* cut to match the 300× *call* cut.

### The staleness arithmetic is safe
- **`MAX_BLOCKS_BEHIND`**: the guard uses `head.saturating_sub(next_block)` for the lag test, and inside the branch `head - max_blocks_behind` can't underflow — the branch condition `head - next_block > max_blocks_behind` with `next_block ≥ 0` guarantees `head > max_blocks_behind`, so `new_start > next_block ≥ 0` and the logged skip count is positive. The subsequent `head < next_block` check then correctly proceeds rather than sleeping. `0 = never drop` preserves the comprehensive default; Sepolia's 300 is the right place to opt into dropping.
- **`QUEUE_JOB_TTL_SECS`**: `age_secs` returning `None` for `enqueued_at <= 0` (serde-default legacy payloads) is the correct backward-compat — old jobs never expire and drain out — and it's tested from both sides (`legacy_payload_without_enqueued_at_never_expires`, `age_is_measured_from_first_enqueue`). Clock skew (now < enqueued_at) yields a negative age that simply isn't `> ttl`, so a future-stamped job isn't dropped — fine.
- The clap `true`/`false`-literal gotcha is handled by always setting compose's default to the literal `false` (`HEURISTIC_ONLY=${HEURISTIC_ONLY:-false}`), which parses correctly rather than "present ⇒ true." Good that you verified that empirically.

### One process question (please confirm before merge)
The lineage note says #170 was merged (`0af7474`) then reverted on main (`50d117f`), and this re-lands the **identical tree**. It doesn't say *why* #170 was reverted. Re-landing an identical tree is only safe if that revert was procedural (a rollback for release timing / unblocking something) rather than a fix for a defect in #170 — otherwise this faithfully re-introduces whatever prompted it. The code itself is validated (green CI against current `main`, which already carries #166, so the "reconciliation with #166" still holds), so this is due diligence, not a code doubt: can you confirm the revert reason was unrelated to a bug in this content? If it was procedural, I'm satisfied.

### Minor (non-blocking)
- **Dropped-block observability.** The worker counts expired drops (`expired_dropped`, logged every 1000), but the head-tracker's forward-skip only emits a `warn` per skip with no running counter/metric. Since the entire purpose of `MAX_BLOCKS_BEHIND` is to *drop data* to bound staleness, a skipped-blocks counter would make the resulting completeness gap visible in the dashboard (how much am I missing?) rather than only inferable from logs. Worth symmetry with the worker's counter.
- **Requeue timestamp preservation.** The TTL-spans-retries guarantee relies on requeues not resetting `enqueued_at`. That path isn't in this diff; the struct-carries-it design supports it, but worth a quick confirm the requeue re-serializes the existing job rather than rebuilding it with a fresh timestamp.
- **`heuristic_only` lives in `CommonConfig`** (shared) though only the worker consumes it — harmless (head-tracker accepts and ignores it) and consistent with `min_gas_used`'s placement, just noting it reads slightly oddly next to `QUEUE_JOB_TTL_SECS` being worker-scoped and `MAX_BLOCKS_BEHIND` head-tracker-scoped.

Solid quota-reduction work with honest tradeoff docs (the call-heavy-tx zero-savings clamp is stated plainly and deferred to #171). Approving — just confirm the revert lineage.

bagelface:
There are direct reference to the issue number #162 in the comments which I think should be stripped out to be more generic.

PR #183, opened 2026-08-13
fix: heuristic accuracy — refunds, SSTORE pricing, calldata, re-entrancy flag

rubydusa:
Fixes the known heuristic discrepancies documented in `crates/core/src/heuristic.rs`, making heuristic-only mode (#182, #162) produce usable numbers.

> **Lineage**: stacked on #182 (the re-land of #170, which was merged in `0af7474` and reverted in `50d117f`). Succeeds #171 — GitHub's stack machinery pinned that PR to the original, now-dead `Rubydusa/162-heuristic-only-mode` branch and rejected every base change, so the PR is recreated here with the identical branch (`Rubydusa/heuristic-gas-refunds`) and content.

## The problem

The heuristic compares pre-refund, mispriced components against post-refund receipt `gasUsed`. On the validated MEV arb tx (`0x19daad7a…a16795`) it overshot by ~19%, clamping reported savings to 0% — which is what heuristic-only mode would have served for essentially every call-heavy DeFi tx.

## The fixes (one commit each)

- **EIP-3529 refunds**: the trace's final refund counter (verified present on live mainnet traces; 5.6k–59.7k in one sampled block) is netted out, capped at gross/5.
- **SSTORE pricing**: actual per-write charge from struct-log `gasCost` (cold 22,100 / warm 100 / dirty 5,000) instead of flat 5,000. `WARM_SSTORE_COST` is gone.
- **Calldata**: EIP-2028 cost over the ABI-encoded state updates the replay ships (~350k gas on the MEV tx; this bias pushed estimates *down*).
- **Fallback double-count**: `extract_operation_counts_from_trace` no longer counts nested SSTOREs/LOGs whose gas is already inside `external_call_gas`.
- **Re-entrant callbacks — flagged, not corrected** (policy decision): a callee calling back into the origin contract (Uniswap swap callbacks; 72/128 depth-1 calls on the MEV tx) still counts as external gas, but `TraceExtract::reentered` is set — detected via a stack peek on CALL-family opcodes at depth ≥ 2, STATICCALL included (the view-re-entry case from #113). Surfaced as a new `reentered` column on `analysis` (auto-migration), a CLI warning, and `reentered` + optional `origin_address` in the wasm API (backward-compatible).

Also: `compute_state_updates` now returns a `TraceExtract` struct instead of a 4-tuple.

## Reconciliation with main (merge commit)

Main moved substantially while this was in flight; the merge commit reconciles the `TraceExtract` refactor with canonical checkpoints (#173), prestate hybrid extraction (#163/#165), and unbounded sim mode (#166):

- `extract_state_updates_hybrid` keeps its `(updates, skipped, call_gas)` tuple; its legacy struct-log arm adapts to the new `compute_state_updates` signature with re-entry detection off — the simulation path never surfaces the flag, only the historical-tx analyzers do.
- `compute_state_updates_canonical` stays tuple-returning: the heuristic aggregates `TraceExtract` carries have no consumer on the canonical path (its fallback estimates are operations-based).
- `UNBOUNDED_COLD_SSTORE_COST`'s compile-time guard against the removed `WARM_SSTORE_COST` is folded into its doc: the heuristic now charges traced actual SSTORE costs, which EIP-2929/2200 cap at exactly this constant, so the gate's bound holds by construction.

## Verification

- Live end-to-end on a mainnet Universal Router swap (`0xc6b65088…7492`): `reentered=true` with origin supplied / `false` without, and heuristic estimate **120,084 vs actual gasUsed 126,564** (~5% under, vs the old systematic overshoot).
- clippy `-D warnings` across all CI configs (workspace, wasm, anvil, cli+anvil); 8 new unit tests (refund cap, SSTORE gas summing, re-entry positive/negative).

## Not included

- `indexer-web` stores `reentered` but doesn't display it yet.
- The wasm dashboard JS must pass `tx.to` as `origin_address` to activate detection there.
- EIP-7623 floor pricing (replays execute enough that standard token pricing applies).
- A depth-1 self-call (the origin contract CALLing its own address) has the same mispricing as a callback but is not flagged — detection starts at depth ≥ 2.


BreadrichEngels (approved):
## Review — heuristic accuracy (refunds, SSTORE pricing, calldata, re-entrancy flag)

**Approve.** This looked like it might touch consensus — it refactors `compute_state_updates` (which the operator extraction path calls) and removes `WARM_SSTORE_COST` (which the #166 gate had a compile-time assert against) — so I verified both against source, and neither is a problem. It's a reporting-accuracy change plus a metadata flag; the signed digest is untouched. CI: `test-wasm`/`test-indexer`/`lint` green, `test-native`/bench still running (I'd expect green — the core changes are additive and covered by the encoder-agreement + new unit tests).

### Consensus safety — verified, the extracted state updates are byte-identical
This is the part I most wanted to confirm, and it holds:
- `compute_state_updates`'s `(trace) → tuple` becomes `(trace, origin) → TraceExtract`, but the **`state_updates` are still produced by the same unchanged `append_state_update_from_struct_log` calls**. The new `TraceExtract` fields (`refund_counter`, `sstore_gas_total`, `reentered`) are pure read-side additions — `refund_counter` reads the final log, `reentered` is a stack-peek flag, `sstore_gas_total` is an accumulator — none of them feed back into `state_updates`.
- On the operator/simulation path, `extract_state_updates_hybrid` calls `compute_state_updates(trace, None)` and destructures exactly the same `(state_updates, skipped_opcodes, call_gas_total)` it did before, ignoring the new fields; `compute_state_updates_canonical` is untouched (still tuple-returning). So the storage updates that feed the digest are byte-for-byte what they were. `origin = None` also means the whole re-entry block is skipped on that path — zero behavior change there.
- The one control-flow change (`else` → `else if state_updates.len() > update_count_before`) only guards the *side bookkeeping* (call-gas and the new sstore-gas accumulation) against an append that returns `None` without adding a row — it can't change which `state_updates` are produced, and if anything it tightens the call-gas tracking. The `total_call_gas must agree between encoders` test still passes, so the call-gas total is unchanged on the covered trace.

**Net: this PR does not change the operator digest.** It's reporting + a `reentered` flag surfaced only to the historical-tx analyzers (CLI/wasm/indexer).

### The #166 gate-guard removal is handled correctly
Removing `const _: () = assert!(UNBOUNDED_COLD_SSTORE_COST > heuristic::WARM_SSTORE_COST)` is *forced* — `WARM_SSTORE_COST` is gone — and the substituted reasoning is sound: the gate's `22_100` is exactly the EIP-2929/2200 cap on a cold zero→nonzero write, and the heuristic now charges each write its *traced actual* cost, which can't exceed that cap, so the gate still never prices a write below the heuristic — by construction, with no shared constant to drift. Importantly the gate's own constant is unchanged, and the #166 `analytic_bound_dominates_measured_apply_cost` property test (untouched here) still validates the gate against revm-measured cost, which is a stronger guard than the drift-assert it replaces. So the gate's upper-bound guarantee is intact.

*Optional hardening:* since the compile-time tripwire is gone, you could pin the "holds by construction" claim at build time with a fresh `const _: () = assert!(UNBOUNDED_COLD_SSTORE_COST == 22_100)` (or `== COLD_SLOAD + SSTORE_SET`) so a future edit to the gate constant fails the build rather than only the property test. The property test already covers it, so this is belt-and-suspenders.

### The accuracy fixes look right and are validated
The four corrections each address a documented bias in the right direction — EIP-3529 refund netting capped at gross/5, per-write SSTORE `gasCost` from the struct log (cold/warm/dirty) instead of flat 5k, EIP-2028 calldata over the shipped ABI encoding, and dropping the nested-SSTORE/LOG double-count already inside `external_call_gas` — and the live check (120,084 vs actual 126,564, ~5% under, versus the old ~19% systematic overshoot that clamped call-heavy DeFi txs to 0%) is exactly the kind of end-to-end validation this needs, backed by 8 unit tests on the refund cap, SSTORE summing, and re-entry positive/negative. Since none of this feeds the digest, empirical + unit coverage is the right bar and it's met.

### Migration + minor notes
- `ALTER TABLE analysis ADD COLUMN reentered BOOLEAN NOT NULL DEFAULT FALSE` is clean and backward-compatible (existing rows default false), and the INSERT binds it as `$19`. One thing to confirm: the migration file is timestamped `20260712…` (from #171's original authoring) but lands now — on a fresh DB (CI) it applies in order, but on the **live indexer DB** verify it doesn't sort *before* an already-applied migration and trip sqlx's out-of-order check. Low stakes (reporting DB), just worth a glance before deploy.
- **Re-entry flag scope** is documented and fine as a *qualify-not-correct* signal: flagging DELEGATECALL/CALLCODE-to-origin (context-preserving, not true re-entrancy) is conservative over-flagging, and the un-flagged depth-1 self-call is a stated limitation — neither affects anything but the advisory column.
- **Lineage**: like #182 this is a recreate (of #171, GitHub stack machinery pinned it to a dead branch) with identical content, stacked on #182 — so the same revert-lineage confirmation still owed on #182 transitively covers this; nothing extra needed here.

Careful, well-scoped work — the refactor is exactly additive where it needed to be, and the digest path is provably unchanged. Approving.

